PR #37, opened 2015-08-14
Add support for saturation family II keywords

totto82:
Saturation functions can now be given using the family II keywords SWFN,
SGFN and SOF3.

Tested on SPE1 with the original deck where the saturation functions are specified using family II keywords. 

Needs PR OPM/opm-parser#556

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp
@@ -2,6 +2,7 @@
 // vi: set et ts=4 sw=4 sts=4:
 /*
   Copyright (C) 2015 by Andreas Lauser
+  Copyright (C) 2015 by IRIS AS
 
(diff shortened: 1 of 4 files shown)


andlaus:
overall this looks very good, I'll merge as soon as OPM/opm-parser#556 is in and you removed that stray `print()` statement. (I haven't checked it the logic is correct, though.)

alfbr:
It is important that we include the modified SPE1 deck as a test. Let us discuss it over the week-end.

andlaus:
> It is important that we include the modified SPE1 deck as a test. Let us discuss it over the week-end.

I agree. opm-data would be the natural place for this I guess.

BTW: are there any news when opm-data will be made public?

totto82:
I used print() to compare the endpoint with the two families, It is removed now.

andlaus:
OPM/opm-parser#556 is in. merging.

PR #42, opened 2015-08-19
Support for Family II saturation functions input

totto82:


totto82:
This still needs testing. It reproduce spe1_new with familyII input if I compare it with opm-core master, but not if I compare it with opm-material's familyI input. 
I also want to add a test for the endpoints.

totto82:
With #41 material familyI match opm-core familyI for spe1_new

PR #57, opened 2015-09-14
Silence warnings

atgeirr:
This should only silence warnings (mostly unused argument), and not make any functional changes.
(Checked that it compiles and runs with SPE9.)


atgeirr:
Will self-merge this since the other maintainer is on vacation.

It will not silence all warnings from this module, but all that are currently triggered by usage from opm-core and opm-autodiff.

PR #58, opened 2015-09-15
Use tabulated oil saturations directly in KROG for familyII

totto82:
Revert bug introduced when refactoring the code in the process of
supporting two-phase simulations. 

Tested on SPE1CASE2_FAMII.DATA


atgeirr:
Thanks!

PR #61, opened 2015-09-21
Buxfix - compressed versus cartesian index.

osae:


blattms:
Great catch @osae. Hopefully this explain some differences between parallel and sequentail runs.

andlaus:
thanks. looks fine and Norne still works. (at least for the first few time steps.) merging.

PR #62, opened 2015-09-21
Table container

joakim-hove:
Followup to: OPM/opm-parser#585

Comments are welcome - but this should not be merged before the opm-parser PR is merged.

Two comments from working on this PR:

1. The tests in opm-material are not based on the BOOST unit test framework; any reason for that? 

2. The opm-material codebase is pure header only; in my opinion that means we must be even stricter with tests. For me the situation here was:
```txt
 1) Make incompatible changes to `opm-parser`.
 2) Build `opm-core` to find where  `opm-material` must be updated.
```

I.e. building `opm-material` gives nothing to verify the even most basic level of correctness in the code. I think that is unfortunate

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp
@@ -183,12 +183,12 @@ struct EclEpsScalingPointsInfo
     {
         // TODO: support for the SOF2/SOF3 keyword family
         auto tables = eclState->getTableManager();
-        const std::vector<SwofTable>& swofTables = tables->getSwofTables();
-        const std::vector<SgofTable>& sgofTables = tables->getSgofTables();
-        const std::vector<SlgofTable>& slgofTables = tables->getSlgofTables();
-        const std::vector<SwfnTable>& swfnTables = tables->getSwfnTables();
-        const std::vector<SgfnTable>& sgfnTables = tables->getSgfnTables();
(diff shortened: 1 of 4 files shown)


opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp
```diff
@@ -199,10 +199,10 @@ struct EclEpsScalingPointsInfo
             Swu = 0.0;
             Swcr = 0.0;
             if (!sgofTables.empty())
-                extractUnscaledSgof_(sgofTables[satRegionIdx]);
+                extractUnscaledSgof_(sgofTables.getTable<SgofTable>(satRegionIdx));
```
andlaus: hm, this seems to be a slightly weird API to me: the type of the table is already implicitly specified by the respective `get*()` method of the `TableManager`, yet the class type needs to be explicitly specified here again. wouldn't it be better to do this directly in the `TableManager`? I mean an API similar to this:

```c++
tableManager.get<SgofTable>(satRegionIdx);
```

joakim-hove:
> hm, this seems to be a slightly weird API to me

Well yes; my long term plan is to to distinguish between table configuration and table instances, a bit like `ParserKeyword` and `DekcKeyword`. At the moment part of the trade-off is trying to make the downstream changes as small and safe as possible.

andlaus:
ok then. merge at your convenience.

joakim-hove:
> ok then. merge at your convenience.

Thank you.

PR #64, opened 2015-09-23
Do not store compressed to cartesian in ecl material law manager

blattms:
It is not needed after initialization. Therefore not storing it prevents some space. Additionally we fix some more warnings.


andlaus:
looks good.

Issue #65, opened 2015-09-25
ECLMaterialLawManager::materialParams_ behaves/is used inconsistently


blattms:
https://github.com/OPM/opm-material/blob/12376bc3d262ffca0401a8477f3c13ba938311af/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp#L210-L228: In the functions `ECLMaterialLawManager::materialLawParam(unsigned elemetIdx)` there is a switch on `ECLMaterialLawManager::hasElementSpecificParameters()`. If `true` the size is assumed to equal to the number elements in the grid, other it is supposed to be the number of `satNumRegions`.

Nevertheless, during `ECLMaterialLawManager::init` the same switch is used to either call `ECLMaterialLawManager::initElemSpecific_` or `ECLMaterialLawManager::initNonElemSpecific_`. But both methods create `ECLMaterialLawManager::materialLawParams_` with the same size as the number of elements and initialize one entry for each element. See https://github.com/OPM/opm-material/blob/12376bc3d262ffca0401a8477f3c13ba938311af/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp#L548-L561 and https://github.com/OPM/opm-material/blob/12376bc3d262ffca0401a8477f3c13ba938311af/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp#L382-L386.

If I am not missing something then this could lead to rather unwanted behaviour during a simulation and should be resolved. I would propose to always create one entry in `ECLMaterialLawManager::materialLawParams_` for each cell and remove the switches from `ECLMaterialLawManager::materialLawParams()` as this seems easier to grasp as varying sizes depending on the properties.

blattms:
@andlaus I have not checked it yet, but this bug could explain the deviations I saw during swatInit in the parallel Norne runs. During it the function `materialLawParams(unsigned elemIdx)` to select the parameters used to calculate the capillary pressure. Due to the distributed grid and the inconsistent initialization the values stored in materialLawParams_[satNumRegions_[elemIdx]]` are different in the two cases as the associated global indexes are different.

andlaus:
> If I am not missing something then this could lead to rather unwanted behaviour during a simulation and should be resolved.

[...]

PR #66, opened 2015-09-25
[bugfix] Fix inconsistent assumptions about the size of materialLawParams_

blattms:
In the functions ECLMaterialLawManager::materialLawParam(unsigned
elemetIdx) there is a switch on
ECLMaterialLawManager::hasElementSpecificParameters(). If true the
size is assumed to equal to the number elements in the grid, other it
is supposed to be the number of satNumRegions.

Note that this function is used when calculating the capillary
pressure in  ECLMaterialLawManager::swatInit.

Nevertheless, during ECLMaterialLawManager::init the same switch is
used to either call ECLMaterialLawManager::initElemSpecific_ or
ECLMaterialLawManager::initNonElemSpecific_. But both methods create
ECLMaterialLawManager::materialLawParams_ with the same size as the
number of elements and initialize one entry for each element.

This commit makes the assumption about the size of
ECLMaterialLawManager::materialLawParams_ in
ECLMaterialLawManager::materialLawParams() the same as in the init
method.

Fixes issue #65 in my opinion. Please double check.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -209,22 +209,14 @@ public:
 
     MaterialLawParams& materialLawParams(unsigned elemIdx)
     {
-        if (hasElementSpecificParameters()) {
-            assert(0 <= elemIdx && elemIdx < materialLawParams_.size());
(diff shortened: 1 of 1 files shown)


andlaus:
I hope this has been fixed more thoroughly by #68. if not, please re-open.

PR #77, opened 2015-10-12
Give access to the shared_ptr of the parameters.

blattms:
This is needed to keep the pointer structure when creating a new material law manager for just a subgrid. For this we either need to be able to copy the stored `shared_ptr` or somehow access the underlying pointer.

An alternative would be to make the version returning a mutable reference private and make stuff in opm-autodiff a friend of EclMaterialLawManager.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -202,14 +202,24 @@ public:
 
     MaterialLawParams& materialLawParams(unsigned elemIdx)
     {
-        assert(0 <= elemIdx && elemIdx < materialLawParams_.size());
-        return *materialLawParams_[elemIdx];
+        return *materialLawParamsPointer(elemIdx);
     }
 
     const MaterialLawParams& materialLawParams(unsigned elemIdx) const
     {
-        assert(0 <= elemIdx && elemIdx < (int) materialLawParams_.size());
-        return *materialLawParams_[elemIdx];
+        return *materialLawParamsPointer(elemIdx);
+    }
+
+    std::shared_ptr<MaterialLawParams>& materialLawParamsPointer(unsigned elemIdx)
+    {
+        assert(0 <= elemIdx && elemIdx <  materialLawParams_.size());
+        return materialLawParams_[elemIdx];
+    }
+
+    const std::shared_ptr<MaterialLawParams>& materialLawParamsPointer(unsigned elemIdx) const
+    {
+        assert(0 <= elemIdx && elemIdx < materialLawParams_.size());
+        return materialLawParams_[elemIdx];
     }
 
     template <class FluidState>
@@ -227,6 +237,15 @@ public:
         return *oilWaterScaledEpsInfoDrainage_[elemIdx];
     }
 
+    const std::shared_ptr<EclEpsScalingPointsInfo<Scalar> >& oilWaterScaledEpsInfoDrainagePointer(unsigned elemIdx) const
+    {
+        return oilWaterScaledEpsInfoDrainage_[elemIdx];
+    }
+
+    std::shared_ptr<EclEpsScalingPointsInfo<Scalar> >& oilWaterScaledEpsInfoDrainagePointer(unsigned elemIdx)
+    {
+        return oilWaterScaledEpsInfoDrainage_[elemIdx];
+    }
 private:
     void readGlobalEpsOptions_(Opm::DeckConstPtr deck, Opm::EclipseStateConstPtr eclState)
     {


andlaus:
I don't really get why you need this. My assumption is this: You want to create a EclMaterialLawManager for the undistributed grid. Then you distribute the grid and want to do the same with the material law parameters?

If the above is correct: is there a reason why don't you create the material law manager *after* the grid was distributed over the processes? (i.e., that each process creates an EclMaterialLawManager object for its "subgrid" from the outset.)

blattms:
> If the above is correct: is there a reason why don't you create the material law manager *after* the grid was distributed over the processes? (i.e., that each process creates an EclMaterialLawManager object for its "subgrid" from the outset.)

I think I told you this 2-3 weeks ago:
Yes, the values are not determined after setting up the
EclMaterialLawManager. They are changed afterwards by some other
property managers. I also tried to set up all properties just for the
distributed  grid but this also fails miserable as there are a lot of
places (in opm-core) that assume that all grid cells are
represented, see https://github.com/OPM/opm-autodiff/issues/490

andlaus:
> I think I told you this 2-3 weeks ago

I know, and I was too stupid to get your point back than. (just as I do not get it now. ;)

> They are changed afterwards by some other property managers

the only instances of this which I can come up with from the top of my head are SWATINIT and the hysteresis updates. but these do not require "non-local" knowledge, i.e., they can be basically expressed by the pseudocode

```python
map(lambda Sw, cellIdx: applySwatinit(params[cellIdx], Sw), initialSolution)
```

> I also tried to set up all properties just for the distributed grid but this also fails miserable as there are a lot of places (in opm-core) that assume that all grid cells are represented, see OPM/opm-autodiff#490 

hm. ok, you need a global view for the EQUIL code? how about setting it up globally, use that global manager to calculate the initial solution, then distributing the grid and then all processes setting up their material law parameters for their slice of the distributed grid?

I suppose that the non-local indices are not needed anymore after the initial solution has been calculated. (if they are, this strongly indicates a bug in the code which assumes them to be present IMO.)

alfbr:
> the only instances of this which I can come up with from the top of my head are SWATINIT and the hysteresis updates. but these do not require "non-local" knowledge

Typically, properties can be altered even during simulation in every way. So from a code design perspective, I am surprised if we can assume that all property alterations only require local knowledge, even after initialisation. Granted I haven't spent the necessary time to check out relevant cases, so no examples to share at this point.

dr-robertk:
Assuming the dependency of local values on arbitrary "global" data is a no no for parallel computations.
I think that should be checked carefully.

andlaus:
> Typically, properties can be altered even during simulation in every way.

that's true, but I cannot see how the can be altered across processor boundaries. (i.e., process x needs to modify a parameter of process y !=x)

> So from a code design perspective, I am surprised if we can assume that all property alterations only require local knowledge, even after initialisation. 

that's simply a property of PDEs: each DOF only exchanges fluxes with its direct neighbors (the DOFs which are in some stencil where the DOF is contained.) if "beaming" is allowed (NNCs are present) things get more complicated, but even then one can think of the NNCs as additional neighbors for DOFs...

(note that this property usually does not hold for integral equations like the rendering equation.)

atgeirr:
> that's true, but I cannot see how the can be altered across processor boundaries. (i.e., process x needs to modify a parameter of process y !=x)

I guess this can happen in the overlap domain (x owns the cell, but y needs the update because it is in y's overlap domain).

I admit I am also a little confused by the discussion here, please figure out the proper course of action in time for the release! By adding this to the release milestone I do not insist that it is merged, but that the issue should be resolved, since it looks to me like a potential showstopper for flow_mpi. If it is not, and you want more time, feel free to remove the milestone tag.

andlaus:
> I guess this can happen in the overlap domain (x owns the cell, but y needs the update because it is in y's overlap domain).

sure, but process y has all information to update its own overlap without communicating with x. (i.e. y can do exactly the same thing on its overlap as x does to its interior.)

> I admit I am also a little confused by the discussion here, please figure out the proper course of action in time for the release! By adding this to the release milestone I do not insist that it is merged, but that the issue should be resolved, since it looks to me like a potential showstopper for flow_mpi. If it is not, and you want more time, feel free to remove the milestone tag.

I'm pretty sure this PR is not needed, but maybe there are issues I'm too stupid to see currently. shall I invest a week or so for a proof of concept using `ebos`?

atgeirr:
> shall I invest a week or so for a proof of concept using ebos?

I hope you are able to work this out between you with a bit less effort than that...

andlaus:
> I hope you are able to work this out between you with a bit less effort than that...

Compared to `flow` it's much less effort since most of eWoms works in parallel already.

blattms:
This is a very interesting discussion. This PR implements rather small
changes and raises a principle discussion about how parallelization
should be done.

Maybe we can save ourselves a lot of time be sticking to the changes
and not to the (seemingly) controversial reason why they could be needed.

@andlaus What are the reason not to make these changes? It seems to me
that even in the current situation the parameters can be changed by
the user, can't they?

BTW as long as I have not figured out how and when these parameters a
changed, I cannot issue a PR in flow and do not need this.

atgeirr:
> BTW as long as I have not figured out how and when these parameters a
> changed, I cannot issue a PR in flow and do not need this

So this is not a show-stopper for the release?

andlaus:
> @andlaus What are the reason not to make these changes? It seems to me that even in the current situation the parameters can be changed by the user, can't they? 

the main reason is that I cannot see how this helps you with properly implementing the parallel stuff, and also that these pointers are supposed to be implementation details (i.e., the flexibility to change them to raw pointers or make them per-value without visible effects to the consumer code should be kept).

That said, if you also open the companion PRs for `opm-autodiff` and `opm-core` which make the simulator work correctly in parallel, I'm open for merging this one. (as of now, I don't really see how this patch alone will change anything for the better.)

> So this is not a show-stopper for the release?

I cannot see a reason why it should be one.

atgeirr:
> I cannot see a reason why it should be one.

I have removed it from the release milestone. If you disagree @blattms, tell us!

andlaus:
ok. seems like you're serious about this, so I'll merge. Note that the fact that ebos now works for MPI parallel simulation means that this is not really required...

PR #78, opened 2015-10-12
Fix the SOF3 input  to make the KROW input similar as in SWOF.

totto82:
With the same input the tests pass without problems.


andlaus:
thanks

PR #88, opened 2015-10-23
Bump version and unify format of dune.module.

atgeirr:
After this, all code modules have the same formatting in dune.module, and the correct version name (for the master branch): "2016.04-pre".


andlaus:
thanks. should an analogous patch go into the release branch?

atgeirr:
> should an analogous patch go into the release branch?

Yes, but Arne Morten is handling that (do not push things to release branches until he has fixed the issues encountered).

PR #103, opened 2016-01-04
Using new table api

joakim-hove:
This PR is a followup to: https://github.com/OPM/opm-parser/pulls

The changes are mainly:
 - For PVTG and PVTO the concepts *inner* and *outer* table are not used - instead *undersaturated* 
and *saturated* respectively are used.
 - Some (ugly) copying from `Opm::TableColumn` to `std::vector<double>`


andlaus:
thanks.

andlaus:
hm, this is no longer mergable (due to conflicts with #102). Except for that, the changes seem to be more or less simple substitutions and are good to go in...

PR #109, opened 2016-01-15
Add tests using Scalar=float (try 2).

dr-robertk:
This PR revised all tests such that Scalar=double and Scalar=float is tested. All tests compile, some of the test do not work with float and are therefore disabled for now.

### DIFF
--- a/opm/material/common/Spline.hpp
+++ b/opm/material/common/Spline.hpp
@@ -706,11 +706,9 @@ public:
     /*!
      * \brief Return true iff the given x is in range [x1, xn].
      */
-    bool applies(Scalar x) const
-    {
-        return x_(0) <= x && x <= x_(numSamples() - 1);
-    }
-
+    template <class Evaluation>
+    bool applies(const Evaluation& x) const
+    { return x_(0) <= x && x <= x_(numSamples() - 1); }
 
     /*!
      * \brief Return the number of (x, y) values.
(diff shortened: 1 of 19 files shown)


andlaus:
ok, looks good. once you have merged my fixes into your branch, I'll merge the result into master.

joakim-hove:
Now - this PR looks fine to me; however I wonder whether the templated scalar is necessary at all? I guess it will mostly be `double`- and the unfamiliar `Scalar` where I was expecting a `double` certainly adds to the cognitive load of reading the code. Is it worth it?

andlaus:
> however I wonder whether the templated scalar is necessary at all? 

it is beneficial IMO and it also is not much effort, so the price/benefit ratio is pretty good, I think.

> I guess it will mostly be double- and the unfamiliar Scalar where I was expecting a double certainly adds to the cognitive load of reading the code. Is it worth it?

in eWoms most simulations for the flash model use quadruple precision floating point types (if the compiler supports them) to get better convergence. Besides that, using single-precision floating point scalars potentially leads to better performance than double precision ones, so I guess it is worth the additional mental load. (Also note that the localized automatic differentiation code heavily depends on the ability of using Scalar != double.)

dr-robertk:
@joakim-hove: some time ago somebody claimed that using single precision would speed up the simulator. I don't think it will, but right now it's not even possible to switch to anything other then double because it's hard coded all over the place. In opm-material we still have the chance to prevent this. This is basically the patch for that.

PR #111, opened 2016-01-21
Fixes missing includes (opm-parser PR-656)

jokva:
Adds missing includes which due to PR-656 in opm-parser are no longer
automatically pulled from other headers.


joakim-hove:
This is completely innocent - I'll merge.

andlaus:
thanks for the PR.

I have a request: next time before merging something to opm-parser which might break the downstream OPM modules, could you please build them? (the emphasis is on _before_)

(I just ask because I was basically busy all morning with getting the s*** to compile again instead of doing real work. I suppose that I was not the only one.)

jokva:
All PRs were built and tested before submission. What broke for you?

andlaus:
opm-material and opm-core. but I see that this was my own mistake. (I was on different local branches for these while opm-parser was on master.)

PR #113, opened 2016-01-26
Improve includes from opm-parser

jokva:
Adopting to opm-parser PR#661, add previously missing includes.

https://github.com/OPM/opm-parser/pull/661


andlaus:
looks harmless enough. thanks.

PR #122, opened 2016-02-10
Change Deck access methods/types to references

jokva:
opm-parser#677 changes the return types for the Deck family of classes.
This patch fixes all broken code from that patch set.

https://github.com/OPM/opm-parser/pull/677


andlaus:
I had a cursory look: no objections. please merge together with OPM/opm-parser#677

PR #129, opened 2016-02-26
Add missing opm-parser headers

jokva:
opm-parser pull #695
https://github.com/OPM/opm-parser/pull/695

Can be merged before upstream.


andlaus:
that patch looks very innocent even without OPM/opm-parser#695 so I merge it right now.

did you check that the ebos simulator (from the ewoms) still compiles?

jokva:
I'll have an extra look, but if it's disabled by default then it hasn't been checked.

andlaus:
> I'll have an extra look, but if it's disabled by default then it hasn't been checked.

it is enabled (and compiled by default) as soon as a working dune-cornerpoint is detected. you need to include eWoms in your build though...

jokva:
It works for me™

andlaus:
> It works for me™

cool! thanks for testing.

PR #130, opened 2016-02-29
Silence unused argument warnings.

atgeirr:
This time using the OPM_UNUSED attribute way.


andlaus:
All good, merging. sorry for letting this linger for so long, but github still seems like it only considers some events important enough to send me a notification email. opening this PR was not among them :(

PR #149, opened 2016-06-10
update jenkins build scripts

akva2:
update build order, opm-output moved in the dependency tree


andlaus:
please merge this yourself, it seems like travis does not like this for soome reason but I consider it possible that this is spurious...

andlaus:
hm. I noticed that the compiler version used by travis is `gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3`. As far as I know GCC 4.6 is unsupported by OPM and can break any time. (opm-material itself supports it at the moment, but opm-parser does not, AFAIK.)

anyway, this PR seems very innocent, so I'm merging it. thanks.

PR #155, opened 2016-07-06
Fix const method overload.

atgeirr:
A `const shared_ptr<Foo>` cannot be converted to a `shared_ptr<const Foo>`.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -218,7 +218,7 @@ public:
         return materialLawParams_[elemIdx];
     }
 
-    std::shared_ptr<const MaterialLawParams>& materialLawParamsPointer(unsigned elemIdx) const
+    const std::shared_ptr<MaterialLawParams>& materialLawParamsPointer(unsigned elemIdx) const
     {
         assert(0 <= elemIdx && elemIdx < materialLawParams_.size());
         return materialLawParams_[elemIdx];


jokva:
This is a **huge** red flag.

atgeirr:
?

andlaus:
> A const shared_ptr<Foo> cannot be converted to a shared_ptr<const Foo>.

really? I thought that was only the case for some oldish compilers. (GCC 4.4 anyone?) where has this issue bitten you?

given that I don't  think that this method is strictly necessary in the first place, I don't mind merging the patch if it does not break anything, but I'd appreciate some testing. (the method was introduced by @blattms in #77 and is apparently required for the MPI version of flow, but I never really understood why it was needed.)

andlaus:

jokva:
```
#include <iostream>                                                                                                                     
#include <memory>

class Foo {
    public:
        void print() {
            std::cout << "Mutable. x = " << this->x++ << std::endl;
        }

        void print() const {
            std::cout << "Immutable. x = " << this->x << std::endl;
        }

    private:
            int x = 0; 
};

std::shared_ptr< Foo > ptr; 

const std::shared_ptr< Foo > getptr() {
    return ptr; 
}

int main() {
    ptr.reset( new Foo() );

    auto p = getptr();
    p->print();
    p->print();
    p->print();
    p->print();
}
```

prints:
> Mutable. x = 0
Mutable. x = 1
Mutable. x = 2
Mutable. x = 3

andlaus:
yeah, I probably misread a bit: I meant that

```c++
#include <memory>
int main()
{
  auto foo = std::make_shared<int>(123);
  auto bar = std::make_shared<const int>(456);
  bar = foo;
  return 0;
}
```

works on modern compilers, but the issue seems to be slightly different.

what I'm wondering about is why this has not a problem until now, i.e. what are you trying to accomplish @atgeirr?

atgeirr:
I may be attributing the wrong cause, in any case this was my error:
```
/Users/atgeirr/opm-build-debug/opm-material/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp:224:16: error: non-const lvalue reference to type 'shared_ptr<const
      EclMultiplexerMaterialParams<[3 * ...]>>' cannot bind to a value of unrelated type 'const shared_ptr<EclMultiplexerMaterialParams<[3 * ...]>>'
        return materialLawParams_[elemIdx];
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I think this overload has never been used before. The mutable version has always been used, even from const contexts such as opm/core/props/satfunc/SaturationPropsFromDeck.cpp line 152.

jokva:
Regardless, the fix is to not return a `shared_ptr` but either a raw pointer or a proper reference.

atgeirr:
Changing
```
const auto& params = materialLawManager_->materialLawParams(cells[i]);
```
to
```
const auto& mlm = *materialLawManager_;
const auto& params = mlm.materialLawParams(cells[i]);
```
will trigger the above compile error.

atgeirr:
> Regardless, the fix is to not return a shared_ptr but either a raw pointer or a proper reference.

I:
 - don't see the harm, but
 - think the method's only used by the materialLawParams() method, that indeed returns a reference, so
 - we could just delete it and redo the materialLawParams() method (the const overload).

andlaus:
> [...]
> will trigger the above compile error.

how can that be? the `materialLawParamsPointer()` method does not seem to be involved there. I'll check this.

jokva:
>  - don't see the harm

You're not sharing ownership (as evident by the reference-to-shared-pointer), so in my opinion this was a bug in the first place.

>  - we could just delete it and redo the materialLawParams() method (the const overload).

Which is what I suggested, no?

GitPaean:
```
const std::shared_ptr< Foo > getptr() {
    return ptr; 
}
```

What `const` actually works here in this example? Any difference with no `const`?

atgeirr:
@andlaus I think you misread my initial comment. In your quote an important const was lost:

This works as you say,
```
#include <memory>
int main()
{
  auto foo = std::make_shared<int>(123);
  auto bar = std::make_shared<const int>(456);
  bar = foo;
  return 0;
}
```

but this does not:
```
#include <memory>
int main()
{
  auto foo = std::make_shared<const int>(123);
  auto bar = std::make_shared<int>(456);
  bar = foo;
  return 0;
}
```

jokva:
> What const actually works here in this example? Any difference with no const?

No, and that's the point. It claims it's const but it's really not.

atgeirr:
> does not seem to be involved there

It's called by the materialLawParams(unsigned int) method.

atgeirr:
> No, and that's the point. It claims it's const but it's really not

@GitPaean You have to be really careful to separate `const sp<Foo>` from `sp<const Foo>` (also with regular pointers for that matter).

andlaus:
crap! when I merged #77 I did not notice the fact that `materialLawParams()` was changed to use `materialLawParamsPointer()`. fix coming.

GitPaean:
> edit: it's worth noting that const shared_ptr doesn't allow you to reset the internal pointer, but you can use it as a mutable path to your pointed-to object. Massive semantic difference.

In that specific example, when using `auto`, `p` is not even a `const shared_ptr`. You can reset it.  I am confused by that.

andlaus:
@atgeirr please merge this one or #156. whichever you like better...

atgeirr:
> You're not sharing ownership (as evident by the reference-to-shared-pointer), so in my opinion this was a bug in the first place.

The code was buggy, yes. It could never compile! Returning a const reference to a shared pointer does not have to be a bug, does it? Not sure what would happen if you tried to take a copy.

> 
> we could just delete it and redo the materialLawParams() method (the const overload).
> Which is what I suggested, no?

Yes, I guess I just said so in a convoluted way.

atgeirr:
A tiny note about your example @jokva:
```
const std::shared_ptr< Foo > getptr() {
    return ptr; 
}
```
I do not get a compiler warning on this, but believe one should (superfluous const) just as one would get from returning a `const int`. Adding const on a by-value return does nothing.

However, the example can easily be modified to show your point, just add a const here:
```
const auto p = getptr();
```

You will still call the mutable method, which I think is surprising to some.

atgeirr:
Better fix in #156.

PR #159, opened 2016-08-08
use ref constructor for EclipseState

pgdr:
The EclipseState object takes a `Deck&`.  The `shared_ptr` constructor will be deprecated in the future.


pgdr:
This is a trivial change in order for material to not break in the future.  I will self-merge it in a couple of days if nobody objects.

andlaus:
> I will self-merge it in a couple of days if nobody objects.

no need for that, merging...

PR #167, opened 2016-09-09
remove the ENDSCALE options check, they're handled by flow.

qilicun:
Since this part has already been handled by opm-simulator, so we should remove this part from opm-material.


andlaus:
sorry for being so late: I somehow missed the the email about the creation of this PR.

In principle, I think that the proper place for such consistency checks is the `EclMaterialLawManager` class. The reason is that the code will screw-up if something's wrong and `flow` is not necessarily the only user of this code.

that said, I'm not very dogmatic on that front and most of the consistency checks are already in the other modules, so I'll merge this PR for now.

PR #170, opened 2016-10-14
The great shared_ptr purge

jokva:
Downstream of https://github.com/OPM/opm-parser/pull/939


andlaus:
all good, merge once necessary.

(I guess this will lead to some minor conflicts with the as of yet unmerged JFUNC PR, but I'll handle these when the time comes...)

jokva:
Likely, as I'm touching a LOT of code. Most changes are -> to . though.

andlaus:
> Most changes are -> to . though.

that's why I'm not worried about these conflicts. go ahead as you like.

PR #175, opened 2016-11-01
Re-implementation of Evaluation.

dr-robertk:
This PR re-implements the Evaluation class to store all internal data in one array, DUNE::FieldVector is used here. This improves the code performance for flow_ebos by about 10%. However, the resulting number of linearizations, linear iterations and newton iterations differ, so this should be handled with care.

### DIFF
--- a/opm/material/densead/Evaluation.hpp
+++ b/opm/material/densead/Evaluation.hpp
@@ -51,19 +51,31 @@ template <class ValueT, int numVars>
 class Evaluation
 {
 public:
+    //! field type
     typedef ValueT ValueType;
 
-    enum { size = numVars };
+    //! number of derivatives
+    static constexpr int size = numVars;
 
-    Evaluation() : value_(), derivatives_()
+protected:
+    //! length of internal data vector
+    static constexpr int length_ = numVars + 1 ;
+
+    //! position index for value
+    static constexpr int valuepos_ = 0;
+    //! start index for derivatives
+    static constexpr int dstart_   = 1;
+    //! end+1 index for derivatives
+    static constexpr int dend_     = length_ ;
+public:
+
+    //! default constructor
+    Evaluation() : data_()
     {}
 
-    // copy other function evaluation
-    Evaluation(const Evaluation& other)
+    //! copy other function evaluation
+    Evaluation(const Evaluation& other) : data_( other.data_ )
     {
-        // copy evaluated function value and derivatives
-        value_ = other.value_;
-        std::copy(other.derivatives_.begin(), other.derivatives_.end(), derivatives_.begin());
     }
 
     // create an evaluation which represents a constant function
@@ -73,26 +85,43 @@ public:
     template <class RhsValueType>
     Evaluation(const RhsValueType& c)
     {
-        value_ = c;
-        std::fill(derivatives_.begin(), derivatives_.end(), 0.0);
+        setValue( c );
+        clearDerivatives();
+        Valgrind::CheckDefined( data_ );
+    }
+
+    // create an evaluation which represents a constant function
+    //
+    // i.e., f(x) = c. this implies an evaluation with the given value and all
+    // derivatives being zero.
+    template <class RhsValueType>
+    Evaluation(const RhsValueType& c, const int varPos)
+    {
+        setValue( c );
+        clearDerivatives();
+        // The variable position must be in represented by the given variable descriptor
+        assert(0 <= varPos && varPos < numVars);
+
+        data_[ varPos + dstart_ ] = 1.0;
+        Valgrind::CheckDefined( data_ );
+    }
+
+    // set all derivatives to zero
+    void clearDerivatives()
+    {
+        for( int i=dstart_; i<dend_; ++i )
+        {
+            data_[ i ] = 0.0;
+        }
     }
 
     // create a function evaluation for a "naked" depending variable (i.e., f(x) = x)
     template <class RhsValueType>
-    static Evaluation createVariable(const RhsValueType& value, unsigned varPos)
+    static Evaluation createVariable(const RhsValueType& value, const int varPos)
     {
-        // The variable position must be in represented by the given variable descriptor
-        assert(0 <= varPos && varPos < size);
-
-        Evaluation result;
-
         // copy function value and set all derivatives to 0, except for the variable
         // which is represented by the value (which is set to 1.0)
-        result.value_ = value;
-        std::fill(result.derivatives_.begin(), result.derivatives_.end(), 0.0);
-        result.derivatives_[varPos] = 1.0;
-
-        return result;
+        return Evaluation( value, varPos );
     }
 
     // "evaluate" a constant function (i.e. a function that does not depend on the set of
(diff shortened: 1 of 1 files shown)


atgeirr:
Good idea, surprised to see that it is 10%!

dr-robertk:
I thinks it's that and the removal of high level functions like std::fill and std::copy which in that case are much better implemented with a simple for loop.

opm/material/densead/Evaluation.hpp
```diff
@@ -168,15 +186,15 @@ class Evaluation
     {
         // while the values are multiplied, the derivatives follow the product rule,
         // i.e., (u*v)' = (v'u + u'v).
-        const ValueType& u = this->value_;
-        const ValueType& v = other.value_;
-        for (unsigned varIdx = 0; varIdx < size; ++varIdx) {
-            const ValueType& uPrime = this->derivatives_[varIdx];
-            const ValueType& vPrime = other.derivatives_[varIdx];
+        ValueType& u = data_[ valuepos_ ];
```
andlaus: `const ValueType&` is sufficient here? bear in mind that copying value type objects can be expensive in the nested case. (for the usual case, i.e. primitive floating point values there should be no difference.)

opm/material/densead/Evaluation.hpp
```diff
@@ -168,15 +186,15 @@ class Evaluation
...
+        const ValueType v = other.value();
```
andlaus: const reference?

opm/material/densead/Evaluation.hpp
```diff
@@ -196,15 +212,15 @@ class Evaluation
     {
         // values are divided, derivatives follow the rule for division, i.e., (u/v)' = (v'u -
         // u'v)/v^2.
-        const ValueType& u = this->value_;
-        const ValueType& v = other.value_;
-        for (unsigned varIdx = 0; varIdx < size; ++varIdx) {
-            const ValueType& uPrime = this->derivatives_[varIdx];
-            const ValueType& vPrime = other.derivatives_[varIdx];
+        ValueType& u = data_[ valuepos_ ];
```
andlaus: const

opm/material/densead/Evaluation.hpp
```diff
@@ -196,15 +212,15 @@ class Evaluation
...
+        const ValueType v = other.value();
```
andlaus: reference

opm/material/densead/Evaluation.hpp
```diff
@@ -196,15 +212,15 @@ class Evaluation
...
+        for (int varIdx = dstart_; varIdx < dend_; ++varIdx) {
+            const ValueType& uPrime = data_[varIdx];
+            const ValueType& vPrime = other.data_[varIdx];
 
-            this->derivatives_[varIdx] = (v*uPrime - u*vPrime)/(v*v);
+            data_[varIdx] = (v*uPrime - u*vPrime)/(v*v);
         }
-        this->value_ /= v;
+        u /= v;
```
andlaus: ah: that's what prevents `u` from being `const`?

opm/material/densead/Evaluation.hpp
```diff
@@ -295,30 +308,27 @@ class Evaluation
...
+    { return data_[ valuepos_ ] == other; }
 
     bool operator==(const Evaluation& other) const
     {
-        if (this->value_ != other.value_)
-            return false;
-
-        for (unsigned varIdx = 0; varIdx < size; ++varIdx)
-            if (this->derivatives_[varIdx] != other.derivatives_[varIdx])
+        for (int varIdx = 0; varIdx <= size; ++varIdx)
```
andlaus: this is pretty non-obvious since `size` is the number of derivatives. better use `data_.size()` or maybe even `return data_ == other.data_;` works?

opm/material/densead/Evaluation.hpp
```diff
@@ -329,53 +339,52 @@ class Evaluation
 
     template <class RhsValueType>
     bool operator>(RhsValueType other) const
-    { return this->value_ > other; }
+    { return data_[ valuepos_ ] > other; }
```
andlaus: mabe it is better to use `value()` in such situations...

opm/material/densead/Evaluation.hpp
```diff
@@ -51,19 +52,28 @@ template <class ValueT, int numVars>
 class Evaluation
 {
 public:
+    //! field type
     typedef ValueT ValueType;
 
-    enum { size = numVars };
+    //! number of derivatives
+    static constexpr int size = numVars;
```
andlaus: hm. possibly the name `size` is a bit confusing. My fault, but better names are appreciated.

opm/material/densead/Evaluation.hpp
```diff
@@ -255,9 +268,9 @@ class Evaluation
     Evaluation operator-() const
     {
         Evaluation result;
-        result.value_ = -this->value_;
-        for (unsigned varIdx = 0; varIdx < size; ++varIdx)
-            result.derivatives_[varIdx] = - this->derivatives_[varIdx];
+        // set value and derivatives to negative
```
andlaus: since these are `Dune::FieldVector`s, is it possible to use `data_ = - other.data_`. (I know, the machine code should be identical, but it would be slightly less clutter here.)

opm/material/densead/Evaluation.hpp
dr-robertk: u is used a bit below. So either const ValueType& u and then data_[ valuepos_ ] below or the way I have put it. I'm usually in favour of keeping the places where valuepos_ is used to a minimum.

opm/material/densead/Evaluation.hpp
dr-robertk: Bu that would be a copy of the value, right? I can check.

opm/material/densead/Evaluation.hpp
```diff
@@ -73,65 +83,76 @@ class Evaluation
...
     }
 
-    // create a function evaluation for a "naked" depending variable (i.e., f(x) = x)
+    // create an evaluation which represents a constant function
+    //
+    // i.e., f(x) = c. this implies an evaluation with the given value and all
+    // derivatives being zero.
     template <class RhsValueType>
-    static Evaluation createVariable(const RhsValueType& value, unsigned varPos)
+    Evaluation(const RhsValueType& c, const int varPos)
```
dr-robertk: What do you mean? Remove createVariable?

andlaus: the additional constructor. it's new, right? (createVariable() stays around AFAICS.)

opm/material/densead/Evaluation.hpp
andlaus: *arg*, both (and in particular the missing `operator==()`) sound awfully like a gap in `Dune::FieldVector`. anyway, thanks for the info.

atgeirr: I added a free (non-class) operator- for FieldVector somewhere long ago (perhaps for CpGrid, don't remember), but that trick would not help much here. It is very annoying. I guess it is to avoid exposing temporary-creating operations, if I remember correctly?

However it prompts me to ask another thing: why not use std::array instead of Dune::FieldVector? It's an invisible internal detail and we don't gain much in terms of ease-of-writing operations for using FieldVector, it seems to me.

opm/material/densead/Evaluation.hpp
atgeirr: I agree, what about simply exposing it as numVars (possibly renaming the template argument numVarsArg or something) or numDerivs?

dr-robertk:
I re-implemented the re-implementation using std::array. It's actually slightly faster than the FieldVector version. I also addressed most of the above comments.

opm/material/densead/Evaluation.hpp
andlaus: > I guess it is to avoid exposing temporary-creating operations, if I remember correctly?

yeah, that's true for `operator-()`, but not for `operator==()`. anyway, I don't think this is important enough to invest (the probably significant) effort to get it into Dune. In particular since we need to work around this issue anyway for several years to come.

dr-robertk:
Has anybody looked at this?

andlaus:
> Has anybody looked at this?

yep. I'm at the final testing. expect it to be merged soon :)

atgeirr:
Before you merge: it was reported to change the results, and I think that should be investigated first, as @dr-robertk suggested in his initial comment. This PR is a performance optimization but is not a requirement for other work to the best of my knowledge so we need not rush this. If only flow_ebos is affected I am inclined to say fine (since it is already producing solutions that are different from flow's) but it was not clear to me if it also affected flow (which uses this class for its property evaluations).

andlaus:
> Before you merge: it was reported to change the results, and I think that should be investigated first, as @dr-robertk suggested in his initial comment.

I looked at this for yesterday's version: the results where the same; convergence was slightly different but the differences were not dramatic and there was no clear picture because for some decks it converged slightly better, for some slightly worse. to me it looked a lot like numerical noise. I'll test it with plain-flow before merge, though.

andlaus:
ok, the normal `flow` simulator produces these results for Norne on my machine:

with this PR:

```

Total time taken (seconds):  1322.9
Solver time (seconds):       1316.25
Overall Linearizations:      1952
Overall Newton Iterations:   1597
Overall Linear Iterations:   25156
```

without this PR:

```

Total time taken (seconds):  1329.96
Solver time (seconds):       1323.25
Overall Linearizations:      1976
Overall Newton Iterations:   1618
Overall Linear Iterations:   25072
```

(`flow_ebos` is at about 750 seconds.) My interpretation of these results is that convergence behaviour is slightly *different*, but overall its *not worse*.  (what can't be inferred from these summaries: the number of time step failures is about the same but they sometimes happen for different time steps.) So if there are no further objections, I'd like to merge this PR tomorrow.

atgeirr:
> So if there are no further objections, I'd like to merge this PR tomorrow.

Seeing as there are significant (but beneficial we hope) changes to flow results, I'd like for this to also be tested on model 2, including in parallel, before we merge this. There are also other changes already in flight that will change results, in PR OPM/opm-simulators#880. I think therefore @totto82 should give his opinion before we merge (any of these will also require updating the regression test references in opm-data).

Finally, I am a bit surprised that this can have such an effect on the results, and I'll remain nervous about that until someone can explain it. A double is a double, or perhaps not?

andlaus:
> I'd like for this to also be tested on model 2, including in parallel, before we merge this.

can you do this? I've never run `flow` in parallel. (since this PR does not affect the parallelization, I'd be *really* surprised if it changed things more for the parallel stuff than for the sequential runs.)

andlaus:
> Finally, I am a bit surprised that this can have such an effect on the results, and I'll remain nervous about that until someone can explain it. A double is a double, or perhaps not?

yeah, but in machine code, a multiplication is not always the same as a multiplication. e.g. the AVX multiplication may round the last bit differently than the x87 variant which may cascade into the differences we see.

jokva:
We attempted a change in the grid cell volume calculations in ert, and the difference between the optimisations before/after were on the scale of 1e-15. Still enough for flow to spit out ~1e-6 difference from its own reference solution, breaking the integration tests.

Safe to say, flow is *very* sensitive to these changes.

andlaus:
> Safe to say, flow is *very* sensitive to these changes.

yeah, for this reason I agree that we should not do them on release branches (or shortly before a release), but I think we should be able to do them at some point. (possibly with adapt the unit tests after looking at them manually and convinving ourselfs that there are no qualitative differences.) BTW: maybe it would be a good idea to add a CPU with a significantly different architecture (ARM?) to the testfarm?

atgeirr:
> can you do this? I've never run flow in parallel.

I can. Might be a little time though before I'm done (days not weeks).

atgeirr:
> we should be able to do them at some point.

Sure, I just want to be sure that the solution paths the simulators end up taking for our important test cases all end well before merging. What I am afraid of is a situation where for example model 2 runs suddenly cannot get past a point in the schedule, due to some bug that might be latent but only triggered by the (chaotic) minor changes.

andlaus:
>  I can. Might be a little time though before I'm done (days not weeks).

great, that sounds like a plan!

totto82:
I have tested this PR with flow on Norne and Model2 and I get exactly the same results for both models.  
My numbers are slightly different than what @andlaus reported.   

```
Total time taken (seconds):  1304.07
Solver time (seconds):       1296.1
Overall Linearizations:      2005
Overall Newton Iterations:   1639
Overall Linear Iterations:   25440
```
Is it possible to ask Travis to check this PR from opm-simulator.

akva2:
travis, no. jenkins, probably. if you explain what 'this PR' means.

totto82:
@akva2 I want to test flow on Norne with PR #175 (from opm-material) and all other repos master.  The only solution I could think of is to make a trivial PR in opm-simulator and then test it with the downstream module opm-material#175. Is there something more elegant?

akva2:
well you can always run it local on your machine. in top of opm-simulators at the state you want to use do

```bash
WORKSPACE=`pwd` BUILDTHREADS=<something suitable> ghprbCommentBody="jenkins build this opm-material=175 please" jenkins/build.sh
configuration=serial WORKSPACE=`pwd` jenkins/run-norne.sh
```

i'll put making the run triggers available from other repos on my TODO.

andlaus:
> Is there something more elegant?

maybe this:

jenkins run Norne opm-material=175 please

@akva2: as far as I know there is no trigger for running Norne with flow_ebos yet?

andlaus:
jenkins run Norne opm-material=175 please

akva2:
when the daddy of the jenkins support says something is not supported, he probably know what he is on about ;) run triggers are only available from opm-simulators and opm-data.

and no, there is no such trigger for flow_ebos. if you want to add it, you have to replicate jenkins/run-norne.sh in opm-simulators plus i have to do some (trivial) scripting on the jenkins side.

atgeirr:
After discussions and testing we have a consensus that this can be merged.

andlaus:
ok. just like @totto82 I also get exactly the same convergence behaviour for Norne with and without this PR  when using `-O2` instead of `-Ofast -ffast-math -fno-finite-math-only` as the compiler flags. merging, thanks to all involved.

(mental note to myself: `-ffast-math` should be called `-fnondeterministic-math`.)

PR #178, opened 2016-11-17
Use tabdims from EclipseState.

atgeirr:
This makes it possible to use the EclMaterialLawManager without explicit TABDIMS in the deck.

Requires OPM/opm-parser#983.

Only very limited testing has been done (other than a full compile): I have verified that this solves my problem with the old case decks without TABDIMS, and that Norne still starts up properly.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -117,8 +117,7 @@ public:
                       const std::vector<int>& compressedToCartesianElemIdx)
     {
         // get the number of saturation regions and the number of cells in the deck
-        int ntsFun = deck.getKeyword("TABDIMS").getRecord(0).getItem("NTSFUN").get<int>(0);
-        unsigned numSatRegions = static_cast<unsigned>(ntsFun);
+        const size_t numSatRegions = eclState.runspec().tabdims().getNumSatTables();
         size_t numCompressedElems = compressedToCartesianElemIdx.size();
 
         // copy the SATNUM grid property. in some cases this is not necessary, but it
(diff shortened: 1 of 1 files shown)


andlaus:
looks good. @joakim-hove: merge with OPM/opm-parser#983, please.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -117,8 +117,7 @@ class EclMaterialLawManager
                       const std::vector<int>& compressedToCartesianElemIdx)
     {
         // get the number of saturation regions and the number of cells in the deck
-        int ntsFun = deck.getKeyword("TABDIMS").getRecord(0).getItem("NTSFUN").get<int>(0);
-        unsigned numSatRegions = static_cast<unsigned>(ntsFun);
+        const size_t numSatRegions = eclState.runspec().tabdims().getNumSatTables();
```
andlaus: to fix the travis complaint, this should be `eclState.runspec().getTabdims()`?

@joakim-hove: is there a reason why the Runspec object is not accessed via `eclState.getRunspec()`?

jokva: Just generally trying to move towards a less noisy interface.

If you look at the `EclipseConfig` (the most recent addition really) it does this. The `get`-prefix'd still exist mostly for compatibility at this point (in my not so humble opinion), and I'd like to both clean up headers, restrict includes and preserve version the parser interfaces in the not-so-distant future.

atgeirr: I don't see a Travis complaint, so to be clear: do I need to change anything? If so, what?

joakim-hove:
Travis has been restarted after merging the parser changes, I assume this can go right in.

andlaus:
> Travis has been restarted after merging the parser changes, I assume this can go right in.

If it's green, press Green :)

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
andlaus: > Just generally trying to move towards a less noisy interface.

I also generally prefer to skip the `get` prefix. That said, IMO it is more important to keep the API conventions somewhat consistent and also I have only a limited desire to do something like "a great `get` purge", so I vote for adding the `get` prefix to the methods that lack it (e.g. `runspec()` -> `getRunspec()`)...

> The get-prefix'd still exist mostly for compatibility at this point (in my not so humble opinion)

IMO, if you care enough to put effort into this, non-`get` variants can be added to everything and the `get` variants of the methods could be kept around for a while but marked deprectated (so that people will get warnings instead of errors).

PR #182, opened 2016-11-28
Use the JFUNC implementation in opm-parser

pgdr:
Use the JFUNC implementation from opm-parser (depends on OPM/opm-parser#970).

The implementation in opm-parser gives the correctly scaled tables.

It's easy to code when there are no tests, or: I have no idea what I'm doing.
@andlaus could you add a test case for the JFUNC scenario that you have?

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
@@ -468,17 +468,7 @@ private:
     template <class Evaluation>
     static Evaluation unscaledToScaledPcnw_(const Params &params, const Evaluation& unscaledPcnw)
     {
-        if (params.config().enableLeverettScaling()) {
-            // HACK: we need to divide the "unscaled capillary pressure" by the
-            // conversion factor for deck to SI pressures because this quantity is
-            // actually *not* a pressure but the (dimensionless) value of the
-            // J-function. Since opm-parser currently always treads this column as a
-            // pressure, it multiplies it with the conversion factor and we have to
-            // revert this here.
-            Scalar alpha = params.scaledPoints().leverettFactor();
-            return unscaledPcnw*(alpha/params.config().deckPressureConversionFactor());
-        }
-        else if (params.config().enablePcScaling()) {
+        if (params.config().enablePcScaling()) {
             Scalar alpha = params.scaledPoints().maxPcnw()/params.unscaledPoints().maxPcnw();
             return unscaledPcnw*alpha;
         }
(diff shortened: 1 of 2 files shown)


andlaus:
> @andlaus could you add a test case for the JFUNC scenario that you have?

you can use Model 2.2 for testing. last time I tried, I had to remove the prediction part from the schedule, though...

opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
```diff
@@ -468,17 +468,7 @@ class EclEpsTwoPhaseLaw : public EffLawT::Traits
...
-            // conversion factor for deck to SI pressures because this quantity is
-            // actually *not* a pressure but the (dimensionless) value of the
-            // J-function. Since opm-parser currently always treads this column as a
-            // pressure, it multiplies it with the conversion factor and we have to
-            // revert this here.
-            Scalar alpha = params.scaledPoints().leverettFactor();
-            return unscaledPcnw*(alpha/params.config().deckPressureConversionFactor());
-        }
-        else if (params.config().enablePcScaling()) {
+        if (params.config().enablePcScaling()) {
```
andlaus: hm: this does not seem to be correct; the J-function is never actually used...

(instead, it is treated as a normal pressure again. I guess this will even throw with OPM/opm-parser#970.)

to fix this, you'll probably have to do something in `EclTwoPhaseMaterial`. what exactly is the non-trivial part ;)

joakim-hove:
> you can use Model 2.2 for testing. last time I tried, I had to remove the prediction part from the schedule, though...

> to fix this, you'll probably have to do something in EclTwoPhaseMaterial. what exactly is the non-trivial part ;)


This is not good enough; if general developers are to be able to contribute to opm-material there *must* be automatic tests. The fact that getting this JFUNC support fully implemented is in my opinion an effective demonstration of the importance of a suite of automatic tests. Right now developers who are not familiar with opm-material will be coding completely in the blind, that is a slow and error prone process.

andlaus:
> This is not good enough; if general developers are to be able to contribute to opm-material there must be automatic tests.

yeah, I see that the situation is unfortunate. Unluckily I'm too stupid to come up with reasonable test data and I am not allowed to extract and publish the stuff which is in Model 2.2. (can you do something about it?)

> The fact that getting this JFUNC support fully implemented has taken so long, is in my opinion an effective demonstration of the importance of a suite of automatic tests.

JFUNC *is* fully implemented since the release. The only thing about it is that the code in opm-material somewhat is a kludge (just as it seems to be in ECL).

alfbr:
I really don't think model 2.2 is what we want to use for a test here. We need to create tailored decks that are really fast to run, otherwise we will be drowning in computational times for all the regression tests. I am away this week, but should be able to help next week.

andlaus:
> I really don't think model 2.2 is what we want to use for a test here.

not Model 2.2, but maybe its satfuncs?

alfbr:
I need to look closely at the functionality before crafting a suitable test. There is very little I can contribute this week though.

joakim-hove:
I have no doubt @alfbr can make very good test data, but that is in my opinion beside the point. Creating and maintaining unit tests which (tries to ....) guarantee that a module works correctly should be an integral part of the developer responsibility.

joakim-hove:
I have no doubt @alfbr can make very good test data, but that is in my opinion beside the point. Creating and maintaining unit tests which (tries to ....) guarantee that a module works correctly should be an integral part of the developer responsibility.

alfbr:
If unit test is what is needed here I agree. On the more complex regression/integration tests, crafting good ones is typically not within developer competency.

PR #191, opened 2017-01-02
Remove unused Deck arguments.

atgeirr:
Silences warnings and simplifies.


atgeirr:
Requires downstream PR OPM/opm-core#1128. When satisfied please merge both simultaneously.

andlaus:
thanks.

PR #192, opened 2017-01-03
Replace Deck usage in PVT init with EclipseState

jokva:
Requires https://github.com/OPM/opm-parser/pull/1019 for `REFs`

### DIFF
--- a/opm/material/fluidsystems/BlackOilFluidSystem.hpp
+++ b/opm/material/fluidsystems/BlackOilFluidSystem.hpp
@@ -180,24 +180,24 @@ public:
      */
     static void initFromDeck(const Deck& deck, const EclipseState& eclState)
     {
-        auto densityKeyword = deck.getKeyword("DENSITY");
-        size_t numRegions = densityKeyword.size();
+        size_t numRegions = eclState.runspec().tabdims().getNumPVTTables();
         initBegin(numRegions);
 
         numActivePhases_ = 0;
         std::fill(&phaseIsActive_[0], &phaseIsActive_[numPhases], false);
 
-        if (deck.hasKeyword("OIL")) {
+        const auto& phases = eclState.runspec().phases();
+        if (phases.active(Phase::OIL)) {
             phaseIsActive_[oilPhaseIdx] = true;
             ++ numActivePhases_;
         }
 
-        if (deck.hasKeyword("GAS")) {
+        if (phases.active(Phase::GAS)) {
             phaseIsActive_[gasPhaseIdx] = true;
             ++ numActivePhases_;
         }
 
-        if (deck.hasKeyword("WATER")) {
+        if (phases.active(Phase::WATER)) {
             phaseIsActive_[waterPhaseIdx] = true;
             ++ numActivePhases_;
         }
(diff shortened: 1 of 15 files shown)


opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -180,56 +180,57 @@ class BlackOil : public BaseFluidSystem<Scalar, BlackOil<Scalar> >
...
             phaseIsActive_[waterPhaseIdx] = true;
             ++ numActivePhases_;
         }
 
         // this fluidsystem only supports two or three phases
         assert(numActivePhases_ >= 2 && numActivePhases_ <= 3);
 
-        setEnableDissolvedGas(deck.hasKeyword("DISGAS"));
-        setEnableVaporizedOil(deck.hasKeyword("VAPOIL"));
+        setEnableDissolvedGas(eclState.cfg().simulation().hasDISGAS());
```
andlaus: I'd not say that The New Way is easier here (and probably also not more fool-proof), but I'm okay with it...

BTW: all-uppercase in function names should be avoid IMO (as should be all-uppercase enums). I agree that it is just a matter of taste (and consistency), though.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
```diff
@@ -59,20 +58,20 @@ class LiveOilPvt
...
         assert(pvtoTables.size() == densityKeyword.size());
 
         size_t numRegions = pvtoTables.size();
         setNumRegions(numRegions);
 
         for (unsigned regionIdx = 0; regionIdx < numRegions; ++ regionIdx) {
-            Scalar rhoRefO = densityKeyword.getRecord(regionIdx).getItem("OIL").getSIDouble(0);
-            Scalar rhoRefG = densityKeyword.getRecord(regionIdx).getItem("GAS").getSIDouble(0);
-            Scalar rhoRefW = densityKeyword.getRecord(regionIdx).getItem("WATER").getSIDouble(0);
+            Scalar rhoRefO = densityKeyword[regionIdx].oil;
```
andlaus: should that not be an accessor method?

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
```diff
@@ -166,9 +165,9 @@ class LiveOilPvt
         }
 
         vapPar2_ = 0.0;
-        if (deck.hasKeyword("VAPPARS")) {
-            const auto& vapParsKeyword = deck.getKeyword("VAPPARS");
-            vapPar2_ = vapParsKeyword.getRecord(0).getItem("OIL_DENSITY_PROPENSITY").template get<double>(0);
+        const auto& oilvap = eclState.getSchedule().getOilVaporizationProperties(0);
```
andlaus: again, I'd argue the New Way is harder to understand: to me it is unclear that `getOilVaporizationProperties()` represents the `VAPPARS` keyword and I would not expect it to be part of the schedule...

opm/material/fluidsystems/BlackOilFluidSystem.hpp
jokva: It's a bit clumsy still, but I'm expecting that to change soon™. Half the reason we're doing this is to actually improve our overall interface, making it comfortable to use and that's harder if we also have to expose our intermediate stage.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
jokva: It's a read-only const table property, it really doesn't have to be.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
jokva: See previous comment - downstream should ideally not have to deal with the fact that it's controlled by VAPPARS at all, that's for parser to deal with.

jokva:
Thanks.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
jokva: This isn't the "new way", i.e. vaporization properties have been a part of eclipsestate for a long time.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
andlaus: > it really doesn't have to be.

sure, but that's an implementation detail. in my experience it was often beneficial to have the flexibility of an accessor method because you sometimes want to change the implementation later (e.g., for performance reasons or to fix bugs).

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
andlaus: > downstream should ideally not have to deal with the fact that it's controlled by VAPPARS at all, that's for parser to deal with.

if the same parameters can be specified in multiple ways (*ehem* hello satfuncs!), I agree. if not this just makes things harder to look up in the documentation, IMO.

> This isn't the "new way", i.e. vaporization properties have been a part of eclipsestate for a long time.

"new in for this code" ;)

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
jokva: For performance reasons there's no reason to believe the struct access is the slower option, or am I completely lost here?

andlaus: yes, performance is not relevant here.

here, you could want to get this number from some other object e.g. to let the user change it at runtime. If you now have multiple places where you store the densities, you may want them to all access the same number.

the point is: it's hard to say in advance what you need in a year and accessor methods provide some degree of flexibility.

jokva:
This PR *does* introduce potential behaviour change (which would be a bug), so Norne should be run and verified before it's merged.

andlaus:
IIRC the opm-parser prerequisite PR has been merged? i'd thus like to get this in soonish, but before pushing Green, travis needs to succeed and Norne and "Model 2" should be verified not to regress in quality and performance. I'll probably do the testing tomorrow.

andlaus:
the opm-parser companion PR has not been merged. closing this one as well, but feel free to resurrect it when its time comes...

PR #193, opened 2017-01-03
Read SCALECRS and ENDSCALE from EclipseState

jokva:
Requires https://github.com/OPM/opm-parser/pull/1020

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
@@ -196,8 +196,9 @@ public:
                       const Opm::EclipseState& eclState,
                       Opm::EclTwoPhaseSystemType twoPhaseSystemType)
     {
+        const auto& endscale = eclState.runspec().endpoint_scaling();
         // find out if endpoint scaling is used in the first place
-        if (!deck.hasKeyword("ENDSCALE")) {
+        if( !endscale ) {
             // it is not used, i.e., just set all enable$Foo attributes to 0 and be done
             // with it.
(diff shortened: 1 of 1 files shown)


opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
```diff
@@ -196,8 +196,9 @@ class EclEpsConfig
                       const Opm::EclipseState& eclState,
                       Opm::EclTwoPhaseSystemType twoPhaseSystemType)
     {
+        const auto& endscale = eclState.runspec().endpoint_scaling();
```
andlaus: isn't the `endpoint_scaling()` method name *quite* inconsistent with the rest of the parser?

jokva: Nothing is consistent in parser, but that'll hopefully be addressed soon.

andlaus: > Nothing is consistent in parse

that's a new function isn't it?

> but that'll hopefully be addressed soon.

I hope this is done in a way which makes the conventions used by opm-parser *more* similar to the other modules, not *less*!?

jokva: Maybe - the goal is a better interface.

It'll surely not happen before Deck is private.

andlaus:
this probably needs a little fixup after OPM/opm-parser#1021? @jokva?

andlaus:
this is superseded by #207...

PR #202, opened 2017-02-07
[WIP] Move assertFinalized_ to base class to avoid code duplication.

dr-robertk:
This PR moves the assertFinalized check to a base class to avoid code duplication and allow for easy toggle via preproc variable. 

Before I change all remaining headers I wanted to confirm that this solution is ok for everybody.

### DIFF
--- a/opm/material/common/AssertFinalized.hpp
+++ b/opm/material/common/AssertFinalized.hpp
@@ -1,0 +1,77 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ * \copydoc Opm::AssertFinalized
+ */
+#ifndef OPM_MATERIAL_ASSERT_FINALIZED_HPP
+#define OPM_MATERIAL_ASSERT_FINALIZED_HPP
+
+#include <type_traits>
+#include <cassert>
+#include <memory>
+
+namespace Opm {
+
+/*!
+ * \brief Default implementation for asserting finalization of parameter objects.
+ *
+ */
+class AssertFinalized
+{
+#if OPM_CHECK_PARAM_FINALIZED
+    bool finalized_;
+#endif
+
+protected:
+    /*!
+     * \brief The default constructor.
+     */
+    AssertFinalized()
+#if OPM_CHECK_PARAM_FINALIZED
+        : finalized_( false )
+#endif
+    {
+    }
+
+    void assertFinalized_() const
+    {
+#if OPM_CHECK_PARAM_FINALIZED
+        assert(finalized_);
+#endif
+    }
+
+public:
+    /*!
+     * \brief Mark the object as finalized.
+     */
+    void finalize()
+    {
+#if OPM_CHECK_PARAM_FINALIZED
+        finalized_ = true;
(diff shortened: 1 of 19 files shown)


andlaus:
to me the approach seems good, so I've got no real objections.

The biggest question I have about this is how to make sure this is tested it if we set the default of the macro to 0. maybe it should default to off, but be explicitly enabled in jenkins.

opm/material/common/AssertFinalized.hpp
```diff
@@ -0,0 +1,77 @@
...
+#if OPM_CHECK_PARAM_FINALIZED
+        : finalized_( false )
+#endif
+    {
+    }
+
+    void assertFinalized_() const
+    {
+#if OPM_CHECK_PARAM_FINALIZED
+        assert(finalized_);
```
andlaus: its better to use `if (!finalized_) OPM_THROW();` here because the attribute will be available even with optimizations enabled.

dr-robertk: Why didn't you do this then in the first place ;-) ? That is of course as simple fix.

opm/material/fluidmatrixinteractions/BrooksCoreyParams.hpp
```diff
@@ -42,19 +43,19 @@ namespace Opm {
  *\see BrooksCorey
  */
 template <class TraitsT>
-class BrooksCoreyParams
+class BrooksCoreyParams : public AssertFinalized
 {
     typedef typename TraitsT::Scalar Scalar;
-
+protected:
+    using AssertFinalized :: assertFinalized_;
```
andlaus: I think I'd prefer to avoid this `using` and use `AssertFinalized::ensure()` (or similar) at the calling sides.

dr-robertk:
@andlaus: I guess OPM_CHECK_PARAM_FINALIZED could be 1 by default until we figure out a bullet proof solution.

opm/material/fluidmatrixinteractions/BrooksCoreyParams.hpp
dr-robertk: What's wrong with using?

dr-robertk:
@andlaus: I have renamed it. That's all I can do for now.

andlaus:
> @andlaus: I guess OPM_CHECK_PARAM_FINALIZED could be 1 by default until we figure out a bullet proof solution.

ok. I'll need to test it on some non-trivial models ("model 2") before merging then. If there is no regression, I think we can also remove that macro entirely.

dr-robertk:
I thinks it should be ok. Then I will change the other headers too.

andlaus:
> Why didn't you do this then in the first place ;-)

because of the reason for this PR: ;)

the attribute vanishes without NDEBUG and so does the assert.

andlaus:
> What's wrong with using?

technically nothing. IMO it looks a bit strange when used for protected attributes of the base class, though...

dr-robertk:
@andlaus: I have adjusted all classes. Please take a careful look.

andlaus:

dr-robertk:
One more thing. I want to preserve the current default state.

dr-robertk:
I have restored the default behavior, i.e. finalize is only used when NDEBUG is defined. However, now it is possible to turn it off independently from NDEBUG.

dr-robertk:

opm/material/common/EnsureFinalized.hpp
```diff
@@ -0,0 +1,88 @@
...
+#define OPM_MATERIAL_ENSURE_FINALIZED_HPP
+
+#include <cassert>
+#include <opm/common/ErrorMacros.hpp>
+
+// TODO: move this variable to config.h
+#define OPM_CHECK_PARAM_FINALIZED 1
+
+#if ! defined(NDEBUG) && OPM_CHECK_PARAM_FINALIZED
+#define USE_OPM_CHECK_PARAM_FINALIZED 1
```
andlaus: it is probably a good idea to only use `OPM_CHECK_PARAM_FINALIZED`, i.e., something like

```c++
#if !defined(NDEBUG) && !defined(OPM_CHECK_PARAM_FINALIZED)
#define OPM_CHECK_PARAM_FINALIZED 1
#endif
```

this would allow to set the `OPM_CHECK_PARAM_FINALIZED` macro via a `-D` compiler flag. maybe I miss something, though...

dr-robertk: It does not really matter right now. I will fix this in a separate PR when opm-common is adapted.

andlaus: ok.

the numbers are in: enabling the check regresses linearization performance for "model 2" by about 2% in optimized mode compared to master and memory requirements increase by about 5 MB (= 0.35%). for the performance I haven't checked whether the 'if()' is to blame or whether it is because of the increased size of the parameter objects.

anyway, IMO this is  good enough for merging once the CI systems are happy...

andlaus:
ok: with both butlers happy,  I'll merge. thanks.

PR #203, opened 2017-02-13
WIP materialLawParamsWellConnection

totto82:
Attempt to implement support for individual relperm for each well connection. 

@andlaus  Is there a test for the current implementation, including hysteresis?

Depends on OPM/opm-parser#1040

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -141,7 +141,7 @@ public:
         for (unsigned satnumIdx = 0; satnumIdx < numSatRegions; ++satnumIdx)
             unscaledEpsInfo_[satnumIdx].extractUnscaled(deck, eclState, satnumIdx);
 
-        initParamsForElements_(deck, eclState, compressedToCartesianElemIdx, satnumRegionArray);
+        initParams_(deck, eclState, compressedToCartesianElemIdx, satnumRegionArray);
     }
 
     /*!
@@ -213,10 +213,22 @@ public:
         return *materialLawParams_[elemIdx];
     }
 
-    std::shared_ptr<MaterialLawParams>& materialLawParamsPointerReferenceHack(unsigned elemIdx)
+    MaterialLawParams& materialLawParamsWellConnection(unsigned elemIdx)
     {
         assert(0 <= elemIdx && elemIdx <  materialLawParams_.size());
-        return materialLawParams_[elemIdx];
+        return *materialLawParams_[elemIdx];
+    }
+
+    const MaterialLawParams& materialLawParamsWellConnection(unsigned int connectionIdx) const
+    {
+        assert(0 <= connectionIdx && connectionIdx <  materialLawParamsForWellConnections_.size());
+        return *materialLawParamsForWellConnections_[connectionIdx];
+    }
+
+    std::shared_ptr<MaterialLawParams>& materialLawParamsPointerReferenceHack(unsigned int connectionIdx)
+    {
+        assert(0 <= connectionIdx && connectionIdx <  materialLawParamsForWellConnections_.size());
+        return materialLawParamsForWellConnections_[connectionIdx];
     }
 
     template <class FluidState>
@@ -316,12 +328,12 @@ private:
         }
     }
 
-    void initParamsForElements_(const Deck& deck, const EclipseState& eclState,
-                                const std::vector<int>& compressedToCartesianElemIdx,
-                                const std::vector<int>& satnumRegionArray)
+    void initParams_(const Deck& deck, const EclipseState& eclState,
+                     const std::vector<int>& compressedToCartesianElemIdx,
+                     const std::vector<int>& satnumRegionArray)
     {
+
         const size_t numSatRegions = eclState.runspec().tabdims().getNumSatTables();
-        unsigned numCompressedElems = static_cast<unsigned>(compressedToCartesianElemIdx.size());
 
         // read the end point scaling configuration. this needs to be done only once per
         // deck.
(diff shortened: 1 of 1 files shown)


opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -346,8 +358,76 @@ class EclMaterialLawManager
...
 
+        // for well connections
+        std::vector<int> satnumRegionArrayConnections;
+        std::vector<int> compressedToCartesianElemIdxConnections;
+
+        const auto& schedule = eclState.getSchedule();
+        const auto& eclGrid = eclState.getInputGrid();
+
+        size_t timeStep = 0;
+        auto wells = schedule.getWells(timeStep); //need all wells ...
```
totto82: Is there a way to get all wells for all times?

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -346,8 +358,76 @@ class EclMaterialLawManager
...
+                int satnumId = completion.getSatTableId();
+                if (satnumId > 0)
+                    satnumRegionArrayConnections.push_back(satnumId);
+                else {
+                    int satnumIdFromCell = satnumRegionArray[eclGrid.activeIndex(cartIdx)];
+                    satnumRegionArrayConnections.push_back(satnumIdFromCell);
+                }
+            }
         }
+        // TODO change satnum/imbnum in epsGridProperties or change in ->extractScaled(...) ?
```
totto82: What endpoints to use in the well?

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -213,10 +213,22 @@ class EclMaterialLawManager
...
 
-    std::shared_ptr<MaterialLawParams>& materialLawParamsPointerReferenceHack(unsigned elemIdx)
+    MaterialLawParams& materialLawParamsWellConnection(unsigned elemIdx)
     {
         assert(0 <= elemIdx && elemIdx <  materialLawParams_.size());
-        return materialLawParams_[elemIdx];
+        return *materialLawParams_[elemIdx];
+    }
+
+    const MaterialLawParams& materialLawParamsWellConnection(unsigned int connectionIdx) const
```
totto82: How to make sure that the connectionIdx maps to the right well Connection Parameter?

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
GitPaean: getWells() ? Not 100% sure though.

andlaus:
> @andlaus Is there a test for the current implementation, including hysteresis?

currently, only the unit tests only tests the API, but not the semantics of that code. I guess this means "no, besides Norne" ;)

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -346,8 +358,76 @@ class EclMaterialLawManager
 
             // read the end point scaling info for the saturation region
             unscaledEpsInfo_[satnumIdx].extractUnscaled(deck, eclState, satnumIdx);
+        }
+
```
andlaus: more extensive comments would be good here, IMO. (partially my fault, I know. sorry.)

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -346,8 +358,76 @@ class EclMaterialLawManager
...
+        EclEpsGridProperties epsGridProperties, epsImbGridProperties;
+        epsGridProperties.initFromDeck(deck, eclState, /*imbibition=*/false);
+        if (enableHysteresis())
+            epsImbGridProperties.initFromDeck(deck, eclState, /*imbibition=*/true);
+
+        initParamsForElements_(materialLawParams_, deck, eclState, compressedToCartesianElemIdx, satnumRegionArray,
```
andlaus: maybe it is better to rename this to `initParamsForReservoir_()` and add a corresponding method for the well connections. I'm fine with the proposed solution, though.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
andlaus: probably none, i.e., use the unscaled values from the tables. Alternatively it could be the ones from the corresponding cell, but this would only make limited sense because the SATNUM/IMBNUM for the connection might be different from that of the cell...

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -346,8 +358,76 @@ class EclMaterialLawManager
...
+    void initParamsForElements_(std::vector<std::shared_ptr<MaterialLawParams> > materialLawParams,
+                                const Deck& deck, const EclipseState& eclState,
+                                const std::vector<int>& compressedToCartesianElemIdx,
+                                const std::vector<int>& satnumRegionArray,
+                                const GasOilScalingPointsVector& gasOilUnscaledPointsVector,
+                                const OilWaterScalingPointsVector& oilWaterUnscaledPointsVector,
+                                const GasOilEffectiveParamVector& gasOilEffectiveParamVector,
+                                const OilWaterEffectiveParamVector& oilWaterEffectiveParamVector,
+                                const EclEpsGridProperties& epsGridProperties,
+                                const EclEpsGridProperties& epsImbGridProperties)
```
andlaus: I don't really like the signature of this method: Remember the "rule of seven"? since it is an internal method, I can live with it if it needs to be, though...

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
andlaus: good question. I guess these must come from the completions of the opm-parser Schedule object when the parameter objects are created. If the completions can change during the simulation, the parameter creation probably needs to be done by a public `updateCompletionParams(schedule, reportStepIdx)`  (or so) method. this method will need to be called by the simulator before each report step. better ideas are appreciated...

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
joakim-hove: Wells are only added - never removed, so `getWells( "last_timestep" )` will give you all the wells, that is the same as no-argument overload: `getWells( )`.

That being said your desire to have all the wells makes skeptical - I fear you are assembling an unneccessary data structure - instead of relying on those already found in the parser. There is way to many instances of: "I also want my version of this information" in the opm codebase.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
joakim-hove: > If the completions can change during the simulation

A completion can change some state during the simulation - but it can not move; that will be a *new* completion.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
totto82: > There is way to many instances of: "I also want my version of this information" in the opm codebase.

I agree. Is there an alternative? The current role of the parser is to provide the necessary input to the simulator. It is one way communication. The parameter file for the relperms can be updated by the simulator due to hysteresis etc. i.e needs two-way communication. But maybe it is desirable to extend the well/completion objects in opm-parser with a data carrier that could store the necessary well/completion parameters and state?

joakim-hove: > Is there an alternative?

I am not sufficiently familar with the current case to answer that - but my general answer is that I think people often should try *a bit harder* before inventing a new datastructure. As I see it an ideal world should have two types of datastructures:

1. Long lived public instances which represent well defined domain quantities - *preferably immutable*. The `Completion` and `Well` objects from the parser are of this type.
2. Short lived value objects, only visible and used within a very limited scope. These can be more tailored to numerical performance/algorithm than the domain.

In particular in the well domain I feel the opm code base has structures which are intermediate between these two; where `struct Wells` and `WellState` are the main offenders - these classes/structures have way too high visibility.

atgeirr: The `struct Wells` is of @joakim-hove's first kind (should be immutable), and `WellState` is of the second kind. The mutable information (for hysteresis etc.) related to each well completion should go in the WellState, and definitely not in the opm-parser Completion class (which is of the first kind, representing "what was given in the deck" without any regard to "what is the current state").

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
totto82: Then adding a id to the completion object in opm-parser seems like a way to go.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
joakim-hove: > The struct Wells is of @joakim-hove's first kind (should be immutable),

OK; but then it should be removed in favor of the `Well` class from opm-parser.

> The mutable information (for hysteresis etc.) related to each well completion should go in the WellState

Sounds right - but I reiterate that the visibility of the *various* `WellState` classes is too high.

totto82:
This PR will be replaced by a new one by @andlaus

PR #204, opened 2017-02-15
Added bubble point and dew point functions

babrodtk:
This adds the required code to get the bubble point and dew point pressures available for output to a restart file

### DIFF
--- a/opm/material/fluidsystems/BlackOilFluidSystem.hpp
+++ b/opm/material/fluidsystems/BlackOilFluidSystem.hpp
@@ -994,6 +994,27 @@ public:
     }
 
     /*!
+     * \brief Returns the bubble point pressure $P_b$ using the current Rs
+     */
+    template <class FluidState, class LhsEval = typename FluidState::Scalar>
+    static LhsEval bubblePointPressure(const FluidState& fluidState,
+                                       unsigned regionIdx)
+    {
+        return saturationPressure(fluidState, oilPhaseIdx, regionIdx);
+    }
+
+
+    /*!
+     * \brief Returns the dew point pressure $P_d$ using the current Rv
+     */
+    template <class FluidState, class LhsEval = typename FluidState::Scalar>
+    static LhsEval dewPointPressure(const FluidState& fluidState,
+                                       unsigned regionIdx)
+    {
+        return saturationPressure(fluidState, gasPhaseIdx, regionIdx);
+    }
+
+    /*!
      * \brief Returns the saturation pressure of a given phase [Pa] depending on its
      *        composition.
      *
--- a/opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
@@ -537,7 +537,10 @@ public:
                 return pSat;
         }
 
-        OPM_THROW(NumericalProblem, "Could find the oil saturation pressure for X_o^G = " << Rs);
+        std::stringstream errlog;
+        errlog << "Could find the oil saturation pressure for X_o^G = " << Rs;
+        OpmLog::problem("liveoil saturationpressure", errlog.str());
+        OPM_THROW_NOLOG(NumericalProblem, errlog.str());
     }
 
 private:
--- a/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
@@ -568,7 +568,10 @@ public:
                 return pSat;
         }
 
-        OPM_THROW(NumericalProblem, "Could find the gas saturation pressure for X_g^O = " << Rv);
+        std::stringstream errlog;
+        errlog << "Could find the gas saturation pressure for X_g^O = " << Rv;
+        OpmLog::problem("wetgas saturationpressure", errlog.str());
+        OPM_THROW_NOLOG(NumericalProblem, errlog.str());
     }
 
 private:


alfbr:
Great! Please also check if there are any performance penalties for doing the calculations.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -994,6 +994,41 @@ class BlackOil : public BaseFluidSystem<Scalar, BlackOil<Scalar> >
...
+    template <class FluidState, class LhsEval = typename FluidState::Scalar>
+    static LhsEval dewPointPressure(const FluidState& fluidState,
+                                       unsigned regionIdx)
+    {
+        assert(0 <= regionIdx && regionIdx <= numRegions());
+
+        typedef Opm::MathToolbox<typename FluidState::Scalar> FsToolbox;
+
+        const auto& T = FsToolbox::template decay<LhsEval>(fluidState.temperature(gasPhaseIdx));
+        const auto& Rs = FsToolbox::template decay<LhsEval>(fluidState.Rs());
```
andlaus: that should be `Rv`? Also, using `getRv_()` is probably a good idea.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -994,6 +994,41 @@ class BlackOil : public BaseFluidSystem<Scalar, BlackOil<Scalar> >
...
+    template <class FluidState, class LhsEval = typename FluidState::Scalar>
+    static LhsEval bubblePointPressure(const FluidState& fluidState,
+                                       unsigned regionIdx)
+    {
+        assert(0 <= regionIdx && regionIdx <= numRegions());
+
+        typedef Opm::MathToolbox<typename FluidState::Scalar> FsToolbox;
+
+        const auto& T = FsToolbox::template decay<LhsEval>(fluidState.temperature(oilPhaseIdx));
+        const auto& Rs = FsToolbox::template decay<LhsEval>(fluidState.Rs());
```
andlaus: it is better to use `getRs_(fluidState, regionIdx)` here (cf, https://github.com/OPM/opm-material/blob/master/opm/material/fluidsystems/BlackOilFluidSystem.hpp#L53 ), because this also works for fluid states that do not feature a `Rs()` method.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -994,6 +994,41 @@ class BlackOil : public BaseFluidSystem<Scalar, BlackOil<Scalar> >
...
+
+        return oilPvt_->saturationPressure(regionIdx, T, Rs);
+    }
+
+
+    /*!
+     * \brief Returns the dew point pressure $P_d$ using the current fluid saturation in the gas phase (Rv)
```
andlaus: there are no saturations involved in these methods as far as I can see?

babrodtk:
Note on performance (before comments from @andlaus). Appears to have no significant impact on performance that I can see, but it would be great if someone else could also double check that. 

With the new output, and simulating the first few steps of Norne with PBPD added to RPTRST
```

Total time taken (seconds):   48.0447
Solver time (seconds):        46.8868
 Assembly time (seconds):     20.5948
 Linear solve time (seconds): 25.3409
 Update time (seconds):       0.581501
 Output write time (seconds): 1.20117
Overall Well Iterations:      42
Overall Linearizations:       92
Overall Newton Iterations:    72
Overall Linear Iterations:    1469


real	0m51.501s
user	0m51.328s
sys	0m0.172s
```
And similarly for master branch:
```
Total time taken (seconds):   48.2409
Solver time (seconds):        47.3196
 Assembly time (seconds):     20.9911
 Linear solve time (seconds): 25.4526
 Update time (seconds):       0.579831
 Output write time (seconds): 0.889668
Overall Well Iterations:      42
Overall Linearizations:       92
Overall Newton Iterations:    72
Overall Linear Iterations:    1469


real	0m51.692s
user	0m51.460s
sys	0m0.228s
```

opm/material/fluidsystems/BlackOilFluidSystem.hpp
babrodtk: Wasn't aware of this one, but I'll change it according to your suggestion

opm/material/fluidsystems/BlackOilFluidSystem.hpp
babrodtk: Yes, that should be Rv, thank you for spotting that mistake! Fix coming up shortly

opm/material/fluidsystems/BlackOilFluidSystem.hpp
babrodtk: I'm not in complete control of this, but isn't it sufficient with the Rs/Rv factors and pressure to determine when the gas/oil bubbles/"dews"?

Perhaps @atgeirr also can comment and see if I've misunderstood something here?

andlaus: > I'm not in complete control of this, but isn't it sufficient with the Rs/Rv factors and pressure to determine when the gas/oil bubbles/"dews"?

that was my point: just don't mention the saturation in the comment ;)

babrodtk:
I've changed the code slightly, replacing 
```THROW(...)``` 
with 
```THROW_NOLOG(...)``` 
and an explicit 
```OpmLog::Problem(tag, text)```
 so that using try-catch from the calling scope in opm-simulators (see sibling pull request OPM/opm-simulators#1056) is possible without spamming the terminal with errors

andlaus:
I've still got no real objections to merging this, just one question: since these two new methods in the fluid system are now trivial, is it beneficial to add them or shouldn't the `saturationPressure()` method be used directly?

babrodtk:
Basically it's a named alias more or less, yes. I think it will be simpler for the next person coming along searching for that specific functionality, so that he/she does not reimplement as I first did.

On 16 Feb 2017, at 17:25, Andreas Lauser <notifications@github.com<mailto:notifications@github.com>> wrote:


I've still got no real objections to merging this, just one question: since these two new methods in the fluid system are now trivial, is it beneficial to add them or shouldn't the saturationPressure() method be used directly?

-

andlaus:
> I think it will be simpler for the next person coming along searching for that specific functionality, so that he/she does not reimplement as I first did.

ok, I see.

andlaus:
It would be nice if you could squash all the patches of this PR together, BTW.

babrodtk:
Commits squashed

andlaus:
cool. merging, what could possiböy go wrong?

PR #209, opened 2017-03-02
Add option to store/restore hysteresis parameters

babrodtk:
This adds the data path for storing the parameters stored for hysteresis, which is required to better handle restarting when hysteresis is enabled.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
+++ b/opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
@@ -148,6 +148,68 @@ public:
         Valgrind::CheckDefined(values[waterPhaseIdx]);
     }
 
+    /*
+     * Hysteresis parameters for oil-water
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    static void oilWaterHysteresisParams(Scalar& pcSwMdc,
+                                         Scalar& krnSwMdc,
+                                         const Params& params)
+    {
+        pcSwMdc = params.oilWaterParams().pcSwMdc();
+        krnSwMdc = params.oilWaterParams().krnSwMdc();
+
+        Valgrind::CheckDefined(pcSwMdc);
+        Valgrind::CheckDefined(krnSwMdc);
+    }
+
+    /*
+     * Hysteresis parameters for oil-water
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    static void setOilWaterHysteresisParams(const Scalar& pcSwMdc,
+                                            const Scalar& krnSwMdc,
+                                            Params& params)
+    {
+        const double krwSw = 2.0; //Should not be used
+        params.oilWaterParams().update(pcSwMdc, krwSw, krnSwMdc);
+    }
+
+    /*
+     * Hysteresis parameters for gas-oil
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    static void gasOilHysteresisParams(Scalar& pcSwMdc,
+                                       Scalar& krnSwMdc,
+                                       const Params& params)
+    {
+        pcSwMdc = params.gasOilParams().pcSwMdc();
+        krnSwMdc = params.gasOilParams().krnSwMdc();
+
+        Valgrind::CheckDefined(pcSwMdc);
+        Valgrind::CheckDefined(krnSwMdc);
+    }
+
+    /*
+     * Hysteresis parameters for gas-oil
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    static void setGasOilHysteresisParams(const Scalar& pcSwMdc,
+                                          const Scalar& krnSwMdc,
+                                          Params& params)
+    {
+        const double krwSw = 2.0; //Should not be used
+        params.gasOilParams().update(pcSwMdc, krwSw, krnSwMdc);
+    }
+
     /*!
      * \brief Capillary pressure between the gas and the non-wetting
      *        liquid (i.e., oil) phase.
--- a/opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
+++ b/opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
@@ -60,6 +60,9 @@ public:
 
     EclHysteresisTwoPhaseLawParams()
     {
+        // These are initialized to two (even though they represent saturations)
+        // to signify that the values are larger than physically possible, and force
+        // using the drainage curve before the first saturation update.
         pcSwMdc_ = 2.0;
         krnSwMdc_ = 2.0;
         // krwSwMdc_ = 2.0;
@@ -152,7 +155,7 @@ public:
     { pcSwMdc_ = value; }
 
     /*!
-     * \brief Set the saturation of the wetting phase where the last switch from the main
+     * \brief Get the saturation of the wetting phase where the last switch from the main
      *        drainage curve to imbibition happend on the capillary pressure curve.
      */
     Scalar pcSwMdc() const
@@ -168,7 +171,7 @@ public:
     //    { krwSwMdc_ = value; };
 
     /*!
-     * \brief Set the saturation of the wetting phase where the last switch from the main
+     * \brief Get the saturation of the wetting phase where the last switch from the main
      *        drainage curve to imbibition happend on the relperm curve for the
      *        wetting phase.
      */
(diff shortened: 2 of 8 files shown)


babrodtk:
@andlaus Thanks for your comment. if I understand you correctly, you would prefer something like 
```
double hyst[2] = { 0.0, 0.0 }; 
const MaterialLawParams& matParams = ebosSimulator().problem().materialLawParams(cellIdx);
matParams.pcSwMdc(hyst);
```
instead of 
```
double hyst[2] = { 0.0, 0.0 }; 
const MaterialLawParams& matParams = ebosSimulator().problem().materialLawParams(cellIdx);
MaterialLaw::pcSwMdc(hyst, matParams);
```
is that correct? The reason that it's done like this is because of the multiplexer class, which performs the static casts into the "real" parameter type. If you have a suggestion on how to better address this, I'm all ears :)

andlaus:
> is that correct?

almost: I'd like to have something like `params.serialize(container)` and `params.deserialize(container)`. The point is that, if I'm not missing something, the multiplexer machinery is not necessary for parameter objects because they can use dynamic polymorphism -- at least if the `(de)serialize()` methods do not need template parameters. (which is the main reason for the multiplexer stuff, BTW.)

atgeirr:
> I'd like to have something like params.serialize(container) and params.deserialize(container)

Seems to me you are proposing writing the parameters to a flat container (array or vector)? I think we really need to know what the individual numbers actually are, since we'll write them to separate named fields in the restart file.

andlaus:
> I think we really need to know what the individual numbers actually are

I see. in that case, I'd prefer to add `oilWaterHystereisParams(cellIdx)` and `gasOilHystereisParams(cellIdx)` methods to EclMaterialLawManager. To me, the semantics of these are unclear if hysteresis is disabled, but the same applies to the current PR as far as I can see?

babrodtk:
> I'd prefer to add oilWaterHystereisParams(cellIdx) and gasOilHystereisParams(cellIdx) methods to EclMaterialLawManager. 

OK, I'll try to make that happen shortly

> the semantics of these are unclear if hysteresis is disabled, but the same applies to the current PR as far as I can see?

Yes,

andlaus:
cool thanks.

just to be clear about this: Although I think that this PR is IMO currently not good enough so that I'd take responsibility for merging, if you or someone else disagrees, you may merge as far as I'm concerned.

atgeirr:
@andlaus this should be ready, could you have a look at it and see if you got it the way you wanted?

opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
```diff
@@ -170,30 +173,36 @@ class EclDefaultMaterial : public TraitsT
...
      */
     template <class Evaluation>
     static void setOilWaterHysteresisParams(const Evaluation& pcSwMdc,
                                             const Evaluation& krnSwMdc,
+                                            const Evaluation& deltaSwImbKrn,
                                             Params& params)
     {
         params.oilWaterParams().setPcSwMdc(pcSwMdc);
         params.oilWaterParams().setKrnSwMdc(krnSwMdc);
+        params.oilWaterParams().setDeltaSwImbKrn(deltaSwImbKrn);
```
andlaus: I guess this is not really necessary because is recalculated if you use `params.oilWaterParams().update()` instead of the individual setter methods. You will need the reversal saturation for the relperm of the wetting fluid, though. the reason why the saturation delta is separately stored by the material parameter objects is that inverting the relperm curve is quite expensive. (since it's been a while since I wrote this code, take that statement with a grain of salt.)

opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
```diff
@@ -170,30 +173,36 @@ class EclDefaultMaterial : public TraitsT
...
      */
     template <class Evaluation>
     static void gasOilHysteresisParams(Evaluation& pcSwMdc,
                                        Evaluation& krnSwMdc,
+                                       Evaluation& deltaSwImbKrn,
                                        const Params& params)
     {
         pcSwMdc = params.gasOilParams().pcSwMdc();
         krnSwMdc = params.gasOilParams().krnSwMdc();
+        deltaSwImbKrn = params.gasOilParams().deltaSwImbKrn();
```
andlaus: same here and a few times below.

babrodtk:
> I guess this is not really necessary because is recalculated if you use ```params.oilWaterParams().update()``` instead of the individual setter methods.

I trust that you are correct about this, but I cannot easily verify it by just looking at the code. By checking it appears that both approaches give the same result on Norne. Will push an update shortly to accommodate your suggestion.

andlaus:
> I cannot easily verify it by just looking at the code

yeah, I'm the first to admit that the code for the ECL satfuncs is quite a beast and that it could use some cleanups and better documentation. anyway, thanks for your work.

opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
```diff
@@ -286,9 +289,9 @@ class EclHysteresisTwoPhaseLawParams : public EnsureFinalized
         Scalar SwKrnMdcImbibition = EffLawT::twoPhaseSatKrnInv(imbibitionParams(), krnMdcDrainage);
         deltaSwImbKrn_ = SwKrnMdcImbibition - krnSwMdc_;
 
-        Scalar pcMdcDrainage = EffLawT::twoPhaseSatPcnw(drainageParams(), pcSwMdc_);
-        Scalar SwPcMdcImbibition = EffLawT::twoPhaseSatPcnwInv(imbibitionParams(), pcMdcDrainage);
-        deltaSwImbPc_ = SwPcMdcImbibition - pcSwMdc_;
```
andlaus: are you sure that it is correct to remove this? this will effectively imply that the drainage curve is always used as the capillary pressure curve. (not sure if pc-hysteresis implemented yet, though. if not, a comment would be good.)

opm/material/fluidmatrixinteractions/EclTwoPhaseMaterial.hpp
```diff
@@ -163,6 +163,72 @@ class EclTwoPhaseMaterial : public TraitsT
...
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    template <class Evaluation>
+    static void setOilWaterHysteresisParams(const Evaluation& pcSwMdc,
+                                            const Evaluation& krnSwMdc,
+                                            Params& params)
+    {
+        const double krwSw = 2.0; //Should not be used
```
andlaus: use `Scalar` instead of `double`. also this function will break if `Evaluation != double`. (but Scalars instead of Evaluations are sufficient as arguments for this method anyway as far as I can see.)

opm/material/fluidmatrixinteractions/EclMultiplexerMaterial.hpp
```diff
@@ -159,6 +159,142 @@ class EclMultiplexerMaterial : public TraitsT
...
+     * Hysteresis parameters for oil-water
+     * @see EclHysteresisTwoPhaseLawParams::pcSwMdc(...)
+     * @see EclHysteresisTwoPhaseLawParams::krnSwMdc(...)
+     * \param params Parameters
+     */
+    template <class Evaluation>
+    static void oilWaterHysteresisParams(Evaluation& pcSwMdc,
+                                         Evaluation& krnSwMdc,
+                                         const Params& params)
+    {
```
andlaus: is this needed? if yes, can't it be moved to the parameter objects?

opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
```diff
@@ -311,16 +314,16 @@ class EclHysteresisTwoPhaseLawParams : public EnsureFinalized
     // largest wettinging phase saturation which is on the main-drainage curve. These are
     // three different values because the sourounding code can choose to use different
     // definitions for the saturations for different quantities
-//    Scalar krwSwMdc_;
+    //Scalar krwSwMdc_;
     Scalar krnSwMdc_;
     Scalar pcSwMdc_;
```
andlaus: `pcSwMdc_` is probably also not required if `deltaSwImbPc_` is unused...

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -242,6 +242,54 @@ class EclMaterialLawManager
         MaterialLaw::updateHysteresis(*threePhaseParams, fluidState);
     }
 
+    template <class Evaluation>
```
andlaus: I'm pretty sure that none of the parameter classes needs to deal with AD. (hence only `Scalar` but no `Evaluation` parameters should be required.)

andlaus:
that's better. can you also remove the remaining `Evaluation` template parameters from the parameter classes?

babrodtk:
Thanks for spotting that, there were a couple of files that weren't included in the commit. Updated now

andlaus:
thanks, but did you forget a few occasions? e.g. the `oilWaterHysteresisParams()` of the `EclMaterialLawManager` does not need it either if I see this correctly. (I guess it is best if you look at the aggregate diff of the PR and search for `Evaluation`.) once these occurrences have been removed and the PR is squashed together, I'll merge.

babrodtk:
@andlaus urk, thanks for spotting. Squashed together, and I think it's Evaluation-free now. :)

andlaus:
that's correct, thanks for fixing it. let's get it in now!

PR #211, opened 2017-03-09
[Discussion] handle a few NaN situation when evaluating saturationPressure()

GitPaean:
With model 2 (or with some other models with testing branch), when outputting the bubble/dew-point pressure, the evaluation of saturation pressure easily encounters assert failure. The reason is that some NaN values are generated during the calculation. 

This PR tries to catch the NaN values generated during the process and avoids to halt the whole simulation due to assert failures without investigate deeper causes. 

Feel free to close it if it is really not the way the thing should be done.   I am really not very familiar with this part and the output module.

Please suggest if it is still workable.

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
@@ -548,6 +548,14 @@ public:
 
         // use the tabulated saturation pressure function to get a pretty good initial value
         Evaluation pSat = saturationPressure_[regionIdx].eval(Rv, /*extrapolate=*/true);
+
+        if ( Opm::MathToolbox<Evaluation>::isnan(pSat) ) {
+            std::stringstream errlog;
+            errlog << "Find initial gas saturation pressure with NaN value for X_g^O = " << Rv;
+            OpmLog::problem("wetgas NaN initial saturationpressure", errlog.str());
+            OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+        }
+
         const Evaluation& eps = pSat*1e-11;
 
         // Newton method to do the remaining work. If the initial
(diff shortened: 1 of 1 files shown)


opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
```diff
@@ -558,6 +566,14 @@ class WetGasPvt
             const Evaluation& fPrime = ((saturatedOilVaporizationFactor(regionIdx, temperature, pSat + eps) - Rv) - f)/eps;
 
             const Evaluation& delta = f/fPrime;
+
+            if ( Opm::MathToolbox<Evaluation>::isnan(delta) ) {
```
andlaus: You can use `Opm::isnan(delta)` or better `!Opm::isfinte(delta)` instead of this. the first just an alias, the second variant also catches infinite values.

andlaus:
the patch is fine, but before merging I'd like to find out what the root cause of this is.

(also, doesn't that exception need to be caught somewhere in the simulator or opm-output?)

andlaus:
ok: I've merged this PR into #212 and resolved the conflicts. I'm closing this one for now, thanks for the patch!

(we still need the downstream code to catch that exception, right?)

GitPaean:
It will be caught in `getSimulatorData` in the model.

andlaus:
is that PR already open?

GitPaean:
I think we do not need a new PR there. `NumericalProblem` is already caught in the function `getSimulatorData()` in the `BlackoilModelEbos`.

andlaus:
cool. even better. (I was deluded because I've reproduced the problem by evaluating the saturation pressures for R=0 at the beginning of linearizations, so if it throws there, chopping time step size is obviously the correct behaviour. sorry for getting this wrong.)

PR #213, opened 2017-03-14
[WIP] Evaluation: specialize evaluations for used numbers in Blackoil setting

dr-robertk:
This is a work-in-progress PR which demonstrates possible performance improvements for flow_ebos. originally brought up by @andlaus. When specializing the the Evaluation class for a given number of numVars the linearization time in Norne can be decreased by about about 20% which leads to an overall performance improvement of about 10%.

Before settling on this version I have tested @andlaus  AVX specialization, template meta programming for loop unrolling and various other things. This seems to be the best solution. However, there are various things that need to be sorted out before merging, e.g. I have only tested with gcc 4.9.4, the specializations should be auto generated etc.


opm/material/densead/Evaluation.hpp
```diff
@@ -47,26 +48,26 @@ namespace DenseAd {
  * \brief Represents a function evaluation and its derivatives w.r.t. a fixed set of
  *        variables.
  */
-template <class ValueT, unsigned numVars>
```
andlaus: is there a particular reason for the `unsigned` -> `int` changes?

opm/material/densead/Evaluation.hpp
```diff
@@ -109,19 +110,33 @@ class Evaluation
...
     {
         // copy function value and set all derivatives to 0, except for the variable
         // which is represented by the value (which is set to 1.0)
         return Evaluation( value, varPos );
     }
 
+
+    // create a function evaluation for a "naked" depending variable (i.e., f(x) = x)
+    template <class RhsValueType>
+    static Evaluation devide(const RhsValueType& a, const Evaluation& b )
```
andlaus: typo? `divide()`

dr-robertk: Ja, typo.

opm/material/densead/Evaluation.hpp
```diff
@@ -239,11 +249,8 @@ class Evaluation
     Evaluation& operator/=(const RhsValueType& other)
     {
         // values and derivatives are divided
-        ValueType factor = (1.0/other);
-        for (unsigned idx = 0; idx < length_; ++idx)
-            data_[idx] *= factor;
-
-        return *this;
+        const ValueType factor = (1.0/other);
+        return this->operator *=( factor );
```
andlaus: can you remove the whitespace in `operator*()`. At first, I was a bit confused by this.

opm/material/densead/Evaluation.hpp
dr-robertk: My experience is that gcc can optimize nothing better than int. And then unsigned is also unspecific. So what is the problem with int? I don't get any warnings and I always use -Wall as in all warnings.

dr-robertk:
I forgot to mention that I wanted to check this with gcc 6.3 but the boost dependency made it impossible to compile opm-parser with a different compiler then the one use for boost :-(

opm/material/densead/Evaluation1.hpp
```diff
@@ -0,0 +1,416 @@
...
+#include <algorithm>
+
+namespace Opm {
+namespace DenseAd {
+/*!
+ * \brief Represents a function evaluation and its derivatives w.r.t. a fixed set of
+ *        variables.
+ */
+template <class ValueT>
+class Evaluation< ValueT, 1 >
```
andlaus: you do not need to specialize the whole class here, it is sufficient to specialize the member functions/operators that are required, a.k.a.,

```c++
template <class ValueT>
Evaluation<ValueT, 1>& Evaluation<ValueT, 1>::operator+=(const Evaluation<ValueT, 1>& b)
{
   data[0] += b.data[0];
   data[1] += b.data[1];
   return *this;
}
```

(I tested this on a little test program, so there may be reality may prevent this.) This would have the advantage that the specialized classes are automatically consistent with the unspecialised one.

opm/material/densead/Evaluation12.hpp
```diff
@@ -0,0 +1,507 @@
...
+#include <algorithm>
+
+namespace Opm {
+namespace DenseAd {
+/*!
+ * \brief Represents a function evaluation and its derivatives w.r.t. a fixed set of
+ *        variables.
+ */
+template <class ValueT>
+class Evaluation< ValueT, 12 >
```
andlaus: whut? 12 derivatives? where?

tests/test_densead.cpp
```diff
@@ -50,9 +50,12 @@
 #include <cassert>
 #include <stdexcept>
 
-static const int numVars = 3;
+//static const int numVars = 3;
+
+template <class Scalar, int numVars>
+struct TestEnv
```
andlaus: what is this class required for?

opm/material/densead/Evaluation.hpp
andlaus: if I remember correctly, these warnings do not occur with `-Wall` but only with `-Wextra` -- or on clang also with `-Weverything`. I don't see them as a problem either (originally it was `int` IIRC), but @atgeirr probably begs to differ ;)

atgeirr:
Nice gain! Is the unrolling of loops the primary reason for it? (If so it is a little surprising, I'd think that compile-time sized loops would be unrolled by the compiler. I guess that would mean the good(?) old technique of template recursion still has its uses.)

andlaus:
> Is the unrolling of loops the primary reason for it?

I guess the reason is that for hand-unrolled loops, "older" c++ compilers (i.e., GCC < 7) are smart enough to emit SIMD instructions. Since I've also seen a ~20% linearization performance improvement on GCC 7, it may be that it was caused by GCC 7 being smart enough to use SIMD for "non-handunrolled" loops.

> I guess that would mean the good(?) old technique of template recursion still has its uses.)

I think it is far from certain that the compiler is smart enough to use SIMD instructions in this case.

dr-robertk:
@andlaus Would be great if you could double check this with gcc 7 just to be sure that this is something we should do or not.

dr-robertk:
@andlaus: At least at my computer there was also not difference between using the AVX specialization and the ones that I have put up besides that my version is compilable on all different kinds of architectures.

dr-robertk:
@atgeirr I also thought that by now compilers would be smart enough to unroll loops etc. That is why I would like to double check this before we merge.

andlaus:
> @andlaus Would be great if you could double check this with gcc 7 just to be sure that this is something we should do or not.

ok, just finished doing this. it seems like GCC 7 is still an improvement:

with this patch:

```

Total time taken (seconds):   668.709
Solver time (seconds):        656.388
 Assembly time (seconds):     258.229
 Linear solve time (seconds): 331.604
 Update time (seconds):       10.4546
 Output write time (seconds): 11.4159
Overall Well Iterations:      1192
Overall Linearizations:       2049
Overall Newton Iterations:    1705
Overall Linear Iterations:    25618
```

without it:

```

Total time taken (seconds):   718.217
Solver time (seconds):        701.665
 Assembly time (seconds):     296.41
 Linear solve time (seconds): 328.48
 Update time (seconds):       10.4254
 Output write time (seconds): 17.1805
Overall Well Iterations:      1193
Overall Linearizations:       2055
Overall Newton Iterations:    1710
Overall Linear Iterations:    25304
```

so the current water mark for Norne stands at about 11 minutes on my machine. for GCC 5.4.0 the impact of this PR was larger, but `flow_ebos` was slower than when using GCC 7 (~690s with the patch vs ~780s without it; unfortunately I was stupid enough to `git clean` these two files away, so I cannot give you more details ATM.)

dr-robertk:
I saw similar results with gcc 6.3 in a different code. So this means we should find a good way on how to construct these specializations.

opm/material/densead/Evaluation.hpp
atgeirr: Well, yes, but then I prefer int, so I am probably willing to fix any warning fallout from this...

atgeirr:
> So this means we should find a good way on how to construct these specializations.

For such specialized things, macros *can* be acceptable, if we cannot find a better way.

atgeirr:
> ok, just finished doing this. it seems like GCC 7 is still an improvement:
> 
> with this patch:
> 
> ================    End of simulation     ===============
> 
> Total time taken (seconds):   668.709
> Solver time (seconds):        656.388
>  Assembly time (seconds):     258.229
>  Linear solve time (seconds): 331.604
>  Update time (seconds):       10.4546
>  Output write time (seconds): 11.4159
> Overall Well Iterations:      1192
> Overall Linearizations:       2049
> Overall Newton Iterations:    1705
> Overall Linear Iterations:    25618
> without it:
> 
> ================    End of simulation     ===============
> 
> Total time taken (seconds):   718.217
> Solver time (seconds):        701.665
>  Assembly time (seconds):     296.41
>  Linear solve time (seconds): 328.48
>  Update time (seconds):       10.4254
>  Output write time (seconds): 17.1805
> Overall Well Iterations:      1193
> Overall Linearizations:       2055
> Overall Newton Iterations:    1710
> Overall Linear Iterations:    25304

I don't get why the iterations etc. changed with this patch, changed instruction orders I assume? Still find the change more significant than expected.

dr-robertk:
That is another reason why we should be careful about this. I haven't done a thorough comparison of results. The speedup did carry over to Model2, btw.

andlaus:
> I don't get why the iterations etc. changed with this patch, changed instruction orders I assume?

that seems very plausible. another explanation could be that the different variants of the same instruction may produce results so that they are not bitwise equivalent, e.g., by using a different rounding scheme for AVX vs x87.  if the difference is below 3%, I nowadays don't loose much thought about them any more and let it sail under the topic "numerical noise". (BTW: when switching compilers, or even the compiler flags, the number of iterations also change slightly.)

alfbr:
Needless to say, this performance gain is very welcome. Anything I can do to help speed this along?

atgeirr:
> Anything I can do to help speed this along?

Extra testing.

Other than that, we must discuss how to get this properly into the code, as @dr-robertk brought up:

> So this means we should find a good way on how to construct these specializations.

What are the options, other than leaving this as is? Did you generate the current code somehow or was it a manual job?

alfbr:
> Extra testing.

Sure, what do you want me to test?

atgeirr:
> Sure, what do you want me to test?

Since you are on a different system than @dr-robertk, starting with verifying Norne and model 2 would be good even if he already tested it. Also: was that old model 2 or one of the new realizations? Then running more of the new realizations would really show us what the gains are across the board, and verify that we don't get pushed into a bad "solution path".

dr-robertk:
@alfbr Running the Models and see if the output is still comparable to what it was before is what I meant. I don't think that this patch breaks anything but one never knows for sure until it's tested.

totto82:
Has this been tested for 2p, solvent and polymer?

atgeirr:
I'll add another question for @dr-robertk in addition to "how did you generate the specializations": has -funroll-loops been on all the time? I am not sure if it is on by default or not under -O2, also that might differ between compilers and compiler versions for all I know. If we can have this improvement "for free" that would be even better...

alfbr:
Guys, it seems this is getting a bit out of proportion. The code is simply about using vectorised intrinsics for the same calculations. You are now asking for testing every property across twelve full field models and various specialised simulators. We are slowly working towards more comprehensive automated testing, but in the mean time we need to keep this practical. Does it suffice if I tested Norne and one realization of Model 2 for all properties?

dr-robertk:
@totto82 No, but it shouldn't make a difference.

totto82:
> @totto82 No, but it shouldn't make a difference.

Great.

atgeirr:
> Does it suffice if I tested Norne and one realization of Model 2 for all properties?

I hope so, I just get nervous when the number of nonlinear iterations changes.

dr-robertk:
@alfbr I don't want to test everybody every Model we have. Just if someone wants to test this patch with his his favourite Model x.y representation that would be great.

totto82:
I can test it for 2p and solvent.

________________________________

alfbr:
Good. @dr-robertk , does it suffice for me to download this PR? What compile switches should I use?

andlaus:
> I'll add another question for @dr-robertk in addition to "how did you generate the specializations"

If I may chip in my worthles 2c: I cannot really see how to use C macros for this. that said, this could be generated using some python script, though. (maybe using some template engine like [Jinja2](http://jinja.pocoo.org/docs/2.9/).) Since build-time auto-generated code tends to cause headaches, I think that we should keep the results in the repo instead of attempting to do it like opm-parser currently does it.

dr-robertk:
@alfbr You can apply this patch and then opm-core and opm-simulators need to be recompiled. I used a variety of flags. Here they are:

-DNDEBUG -Wfatal-errors -fPIC -O3 -funroll-loops -finline-functions -Wall -march=native -mavx2 -ftree-vectorize -fno-stack-protector -mtune=native -pthread --param large-unit-insns=1000 --param inline-unit-growth=1000 --param max-inline-insns-single=1000 --param large-function-growth=1000 --param large-function-insns=1000 -ftree-vectorize 

I wanted to do one more test to see which flags are actually essential.

alfbr:
Thanks. I got impatient, so I just fired off a build. I am using these switches:
-DNDEBUG -O3 -funroll-loops -finline-functions -Wall -march=sandybridge -ftree-vectorize -fno-stack-protector -mtune=native -mavx -pthread

(running on a Sandy Bridge EP system)

andlaus:
> (running on a Sandy Bridge EP system)

a generic way to achive the same is to use `-march=native`. also, using `-Ofast` instead of `-O3` probably produces slightly faster results.

alfbr:
I have now tested with both Norne and Model 2. All summary curves are basically on top of each other, with a few points showing minor deviations. Cell values are also typically very close, with two exceptions. Both RS and SGAS have quite a few cells with significant deviations. However, I see the same level of differences on these two across different builds without this patch, so I conclude that they are simply the arrays which are most sensitive to numerical errors. Conclusively, I believe this patch produce correct enough results.

alfbr:
I have now also been able to look at performance numbers with and without this patch (identical compiler switches as outlined above). The gain is around 6% on both Norne and Model 2. At this point in time (given the already very good performance), this is very good!

atgeirr:

dr-robertk:
@alfbr I guess you could get a bit more when increasing the limits for function inlining. I have the following: 

 --param large-unit-insns=500 \
 --param inline-unit-growth=500 \
 --param max-inline-insns-single=500 \
 --param large-function-growth=500 \
 --param large-function-insns=500 \

It increases the compile time a bit. Also, @andlaus and i have not really settled on the correct way of specializing which routines.

dr-robertk:
The code generation script generating the Evaluation*.hpp files is now in place and from my point of view this is good to be merged.

andlaus:
I think the same. lacking good reasons not to, I'll press green tomorrow. (also, permission to anyone who's trigger happy to press green earlier is hereby granted.)

andlaus:

atgeirr:
I took a look at the python script to see how you did it and found it clever, yet simple. My only worry is that we could get some drift between the main Evaluation template and the specializations whenever we change even minor things in Evaluation. As long as we agree that no change to Evaluation can be made without corresponding changes to the specializations we should be ok. Perhaps it's possible to write tests for that over time?

andlaus:
the travis error is the equil issue that has been annoying us the last few days, the jenkins problem is the SPE1 restart test failing which I think I've seen before and is very likely unrelated to this PR. since Norne works fine for me locally, and the PR yields a significant performance improvement, I'll merge it now. thanks to everyone involved.

> Perhaps it's possible to write tests for that over time?

there already is a [unit test](https://github.com/OPM/opm-material/blob/master/tests/test_densead.cpp) which is quite extensive IMO. (at least I cannot come up with significant gaps out of my head.) If you want  to test additional things there, patches are welcome, as always.

dr-robertk:
@atgeirr: Both, the default Evaluation.hpp and the specializations are generated by the script to avoid deviations. I think that should be fine.

atgeirr:
> there already is a unit test which is quite extensive IMO

I see, my point was not that Evaluation needed more tests as such, but that we needed to make sure the specialized versions are in sync with the general version. Looking at the test it seems this is ok since it is tested with sizes up to 15 which should trigger both versions.

> @atgeirr: Both, the default Evaluation.hpp and the specializations are generated by the script to avoid deviations. I think that should be fine.

I did not realize that it was already that way, but that sounds good and just what I hoped for!

bska:
This breaks the build of all downstream modules that use installed versions of their upstream modules:
```
/data/projects/opm/build/opm/dbg/seq/dune-2.3.1/opm-common/../inst/include/opm/material/densead/Evaluation.hpp:568:41: fatal error: EvaluationSpecializations.hpp: No such file or directory
 #include "EvaluationSpecializations.hpp"
```

We need to **install** `EvaulationSpecializations.hpp` during `cmake --target install`.

dr-robertk:
In that module all headers are included automatically. Could it be that you have to remove your buidl dir? Otherwise I wouldn't know what could be wrong.

bska:
> In that module all headers are included automatically.

You're right.  I spoke too soon.

> Could it be that you have to remove your build dir?

Yes.  This demonstrates the downside to writing `CMakeLists.txt` in terms of globbing operations.  If there were explicit lists of files to build and install then downstream users would not have to manually rerun `cmake`.

dr-robertk:
But they would have to manually modify the CMakeLists_files.txt which sometimes is forgotten too. So either way doom is awaiting you ;-).

bska:
> But they would have to manually modify the `CMakeLists_files.txt` which sometimes is forgotten too

Sure, but the impact of that is **much** smaller.  Failing to update `CMakeLists_files.txt` affects primarily a group of module developers whereas having to manually rerun `cmake` affects both developers and users (including those developers who don't currently work on the module).  The cost/benefit ratio here really does suggest that we use an explicit list of files.

andlaus:
as i see it, developers should always run `cmake .` before they `make all && make install` so this issue is just a little workflow glitch. (alternatively, use `dunecontrol`.) also, I am of the strong opinion that *users* who are not willing to occasionally endure such pains on the bleeding edge should stick to using releases, preferentially via distribution packages.

atgeirr:
Actually this particular issue was decided on for OPM (in favour of the users) a long time ago... I was not really aware that opm-material did not follow the general pattern.

bska:
> developers should always run `cmake .`

Why?  I'm being serious.  It is a completely unnecessary step once the build tree is set up.  It is the job of the build system to ensure that it is always self-consistent.  You are arguing that all users, including those who aren't working on that particular component, always be aware of the intricacies of the build system when it is trivial to ensure that they don't have to be, all for the benefit of the small group of people who are actively working on a piece of software.  Pruning and recreating build trees is a much bigger job when multiplied by the number of users (not necessarily end-users).

> I am of the strong opinion that users  who are not willing to occasionally endure such pains on the bleeding edge should stick to using releases,

Then you are simply mistaken.  Developers should take every step to ensure that experience of fellow users and developers is smooth.

GitPaean:
I do not have the knowledge / opinion about which way is better. I have been always thinking `OPM` building system will trigger `cmake` when trying to `make` whenever `CMakeLists_files.txt` is updated.  But most of time, I do not need to do the `cmake` again.  It is kind of convenient and saving some time.

dr-robertk:
Anyone else wanting to raise his/her opinion?

bska:
> I have been always thinking OPM building system will trigger `cmake` when trying to make whenever `CMakeLists_files.txt` is updated

It does.  That's the whole point of having `CMakeLists_files.txt` in the first place.

> But most of time, I do not need to do the `cmake` again

Correct, because it happens automatically.

andlaus:
> Actually this particular issue was decided on for OPM (in favour of the users) a long time ago... I was not really aware that opm-material did not follow the general pattern.

IIRC it was decided to use these files of lists for `opm-{common,parser,core,simulators}`. `ewoms` and `opm-material` have used globbing from day one, i.e., since quite some time before this issue surfaced the first time. Anyway, given the fact that nobody noticed for about 1.5 years, this is by far not the biggest issue with the build system. (Did anyone count the number of hours lost to sibling builds?)

andlaus:
> Why?

IMO, because the number of instances where the build was broken because someone forgot to update `CMakeLists.txt` is about an order magnitude larger than the number of occasions it broke because of file globbing ;)

andlaus:
> It is the job of the build system to ensure that it is always self-consistent.

true. but this is a build system and reading the words "consistent" and "build system" in the same sentence is weird *g*

andlaus:
> Pruning and recreating build trees is a much bigger job when multiplied by the number of users (not necessarily end-users).

the point here is that if updating `CMakeLists_files.txt` is forgotten, "users" don't even have a way to fix the issue. also running `cmake .` in the build directory before invoking `make` is not terribly complicated and takes about 3 seconds!?

alfbr:
> Anyone else wanting to raise his/her opinion?

Yes. While I have no strong feelings regarding globbing or explicit list (preference on explicit list though), I find it rather saddening that we waste energy energy discussing it again. Is it really so difficult to just go with the flow (pun intended).

dr-robertk:
@alfbr Should we discuss this? ;-)

alfbr:
Only if you serve beer :-)

dr-robertk:
That would be for sure the right setting ;-). We should make a list of these topics for the next OPM meeting: Things to talk about in the coffee breaks O:-).

atgeirr:
> this broke SPE1CASE2 in opm-simulators.

That must be addressed. In retrospect, how could we forget to Jenkins this...? I see the following alternative ways out:
 - Revert this PR until we know why things changed (that is still somewhat suspicious).
 - Do not revert this PR, letting Jenkins keep failing a little longer, while we investigate this.
 - Update the failing test with new data, assuming that (given that we had ok results on integration tests) it is ok that things changed.

@dr-robertk, @andlaus this is mostly on you to decide and fix. I'll help of course if you want me to.

blattms:
IMHO the failing tests are really annoying as one has no possibility to judge whether a PR is broken by the author or by a previously merged  PR, that broke tests.

bska:
> IMHO the failing tests are really annoying as one has no possibility to judge whether a PR is broken by the author or by a previously merged PR, that broke tests.

I agree with @blattms and I'm tempted to say that unless the problem can be easily analysed and corrected within a reasonable time-frame, then this PR (i.e., #213) should be reverted pending a proper solution.

dr-robertk:
@bska, @blattms: Please don't panic. It will be fixed today. One way or the other. Reverting would cause @bska to have to run cmake a second time within a week, and we don't want that discussion again.

joakim-hove:
> IMHO the failing tests are really annoying as one has no possibility to judge whether a PR is broken by the author or by a previously merged PR, that broke tests.

This case illustrates exactly this problem - recursively. Looking back at the travis/jenkins logs from the time of the merge we had a failing `EQUIL` test in opm-core at the time - broken by yours truly. So - this breakage creates problems for other PRs, and another previous PR created problems when this was merged.

Moral of the story: It is extremely valuable to keep the full chain green at all times.

dr-robertk:
@bska: Don't forget to run cmake.

GitPaean:
Hi, @akva2 , are you sure it is this PR broke the test?  A preliminary test does not show this PR changes the curves of the test case `SPE1CASE2` .

atgeirr:
After discussion with @dr-robertk I will revert this PR for now (or rather, create a reverting PR, run that through Jenkins and merge it if green).

I have also read quite carefully through the code, and found no bugs or errors (there were a few minor issues, that I have reported to @dr-robertk for the next iteration of this). Commenting out the explicit inclusion of the specializations and replacing int with unsigned does not restore the tests. So I am at this point a bit stumped as to why results change.

@GitPaean I have tested this and yes, this breaks the test. That will also be double-checked by Jenkins on the revert PR.

GitPaean:
Okay. I post the error message here for information / discussion. 

```
est command: /home/kaib/OPM-master-test/debug/opm-simulators/tests/run-regressionTest.sh "/home/kaib/OPM-master-test/debug/opm-simulators/../opm-data/spe1" "/home/kaib/OPM-master-test/debug/opm-simulators-build/tests/results/flow+spe1" "/home/kaib/OPM-master-test/debug/opm-simulators-build/bin" "SPE1CASE2" "2e-2" "1e-5" "/home/kaib/OPM-master-test/debug/opm-output-build/bin/compareSummary" "/home/kaib/OPM-master-test/debug/opm-output-build/bin/compareECL" "flow" "/home/kaib/OPM-master-test/debug/opm-simulators/../opm-data/spe1/SPE1CASE2.DATA"

 === Executing comparison for summary file ===
17: Comparing '/home/kaib/OPM-master-test/debug/opm-simulators-build/tests/results/flow+spe1/SPE1CASE2' to '/home/kaib/OPM-master-test/debug/opm-simulators/../opm-data/spe1/opm-simulation-reference/SPE1CASE2'.
17: Comparing 122 steps.
17: For keyword BPR:1
17: (days, reference value) and (days, check value) = (730, 6743.78) and (730, 6743.85)
17: The absolute deviation is 0.0688477. The tolerance limit is 0.02
17: The relative deviation is 1.0209e-05. The tolerance limit is 1e-05
17: Program threw an exception: [/home/kaib/OPM-master-test/debug/opm-output/opm/test_util/summaryRegressionTest.cpp:117] Deviation exceed the limit.
17: 13 summary keyword(s) exhibit failures
17: 	 BPR:1
17: 	 	 Fails for: 83 / 122 steps.
17: 		 Largest absolute error: 8.168950e-01
17: 		 Largest relative error: 1.736960e-04
17: 	 BPR:1,1,1
17: 	 	 Fails for: 83 / 122 steps.
17: 		 Largest absolute error: 8.168950e-01
17: 		 Largest relative error: 1.736960e-04
17: 	 BPR:10,10,3
17: 	 	 Fails for: 80 / 122 steps.
17: 		 Largest absolute error: 4.138180e-01
17: 		 Largest relative error: 1.186030e-04
17: 	 BPR:300
17: 	 	 Fails for: 80 / 122 steps.
17: 		 Largest absolute error: 4.138180e-01
17: 		 Largest relative error: 1.186030e-04
17: 	 FGOR
17: 	 	 Fails for: 2 / 122 steps.
17: 		 Largest absolute error: 2.701950e-02
17: 		 Largest relative error: 1.981050e-03
17: 	 FOPR
17: 	 	 Fails for: 71 / 122 steps.
17: 		 Largest absolute error: 1.063570e+01
17: 		 Largest relative error: 1.173270e-03
17: 	 WBHP:INJ
17: 	 	 Fails for: 82 / 122 steps.
17: 		 Largest absolute error: 8.085940e-01
17: 		 Largest relative error: 1.640810e-04
17: 	 WBHP:PROD
17: 	 	 Fails for: 13 / 122 steps.
17: 		 Largest absolute error: 5.705570e-01
17: 		 Largest relative error: 2.792590e-04
17: 	 WGOR:PROD
17: 	 	 Fails for: 2 / 122 steps.
17: 		 Largest absolute error: 2.701950e-02
17: 		 Largest relative error: 1.981050e-03
17: 	 WGPR:PROD
17: 	 	 Fails for: 73 / 122 steps.
17: 		 Largest absolute error: 9.989060e+01
17: 		 Largest relative error: 8.088780e-04
17: 	 WGPT:PROD
17: 	 	 Fails for: 53 / 122 steps.
17: 		 Largest absolute error: 8.512000e+03
17: 		 Largest relative error: 3.728960e-05
17: 	 WOPR:PROD
17: 	 	 Fails for: 71 / 122 steps.
17: 		 Largest absolute error: 1.063570e+01
17: 		 Largest relative error: 1.173270e-03
17: 	 WOPT:PROD
17: 	 	 Fails for: 66 / 122 steps.
17: 		 Largest absolute error: 4.368000e+03
17: 		 Largest relative error: 8.439170e-05
17: === Executing comparison for restart file ===

17: Comparing IWEL...done.
17: Comparing ZWEL...done.
17: Comparing OPM_XWEL...
17: Occurrence in first file    = 2
17: Occurrence in second file   = 2
17: Grid coordinate             = (1, 1, 1)
17: (first value, second value) = (1.52915e+07, 1.52916e+07)
17: 
17: Program threw an exception: [/home/kaib/OPM-master-test/debug/opm-output/opm/test_util/EclFilesComparator.cpp:369] Deviations exceed tolerances.
17: The absolute deviation is 172.13, and the tolerance limit is 0.02.
17: The relative deviation is 1.12565e-05, and the tolerance limit is 1e-05.
17: Comparing '/home/kaib/OPM-master-test/debug/opm-simulators-build/tests/results/flow+spe1/SPE1CASE2' to '/home/kaib/OPM-master-test/debug/opm-simulators/../opm-data/spe1/opm-simulation-reference/SPE1CASE2'.

```

atgeirr:
The reverting PR went through Jenkins and has been merged. This PR can still be used for discussing the matter, at least until a revised PR is available.

PR #216, opened 2017-03-22
EclMultiplexerMaterialParams: use shared_ptr instead of void*.

dr-robertk:
This seems to fix a problem that occurs when this class is copied. However, I advise for careful investigation of whether to use pointers here.


andlaus:
have you tested this for performance on Norne? if there are no regressions, feel free to press green.

dr-robertk:
No, and I also don't think this should be merged. This is simply a fix for making some of @totto82 latest experiments work. I'll let you know when this is done. Btw, void* ....

dr-robertk:
I have now tested this change with Norne and I don't see any change in performance.

dr-robertk:

dr-robertk:

dr-robertk:
This should also go in once the tests succeeded since it fixes a memory problem with the EclMultiplexerMaterialParams class.

andlaus:
the tests succeeded. since there does not seem to be a performance penaltiy, it goes in now. thanks.

PR #217, opened 2017-03-23
Add method to get connectionMaterialLawParams

totto82:
This method changes the saturation table idx in the original material
law parameter object. In the context of ECL reservoir simulators, this
is required to properly handle wells with its own saturation table idx.
In order to reset the saturation table idx in the materialLawparams_
the method needs to be called with the cells satRegionIdx after it has been used for
the well connection.

Hysteresis is not accounted for. Is is a open question how this should
be implemented. A warning is given if hysteresis is combined with a
modified saturation table idx for the connection.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -122,16 +122,16 @@ public:
 
         // copy the SATNUM grid property. in some cases this is not necessary, but it
         // should not require much memory anyway...
-        std::vector<int> satnumRegionArray(numCompressedElems);
+        satnumRegionArray_.resize(numCompressedElems);
         if (eclState.get3DProperties().hasDeckIntGridProperty("SATNUM")) {
             const auto& satnumRawData = eclState.get3DProperties().getIntGridProperty("SATNUM").getData();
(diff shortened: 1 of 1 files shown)


opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -982,13 +1014,18 @@ class EclMaterialLawManager
     std::vector<Opm::EclEpsScalingPointsInfo<Scalar>> unscaledEpsInfo_;
     OilWaterScalingInfoVector oilWaterScaledEpsInfoDrainage_;
 
-    Opm::EclMultiplexerApproach threePhaseApproach_;
+    GasOilScalingPointsVector gasOilUnscaledPointsVector_;
```
atgeirr: If we depend on storing these data (which are cell data, right?), we must probably also fix the parallel case, when the properties gets distributed, to also distribute these data. Otherwise I would expect this to break flow_mpi and also flow_ebos with MPI (eventually).

totto82: They are satnum data, not cell data. If I understands the code correctly all partitions should create and store these vectors independently from the deck.

atgeirr: Ok, then it should be ok. It's only cell based data we need to worry about.

PR #218, opened 2017-03-23
Fixed numerical problems with bubble/dew point pressures

babrodtk:
This addresses the issues with the dew / bubble point pressures causing excessive terminal output.

The issues were multiple:

 * The tables `RsTable` and `RvTables` may have flat regions in them, thus with a zero derivative. Previously it threw an exception + log problem, now it accepts the initial guess/evaluation as the value.

 * The `saturationPressure_` tables may have multiple identical `x` (Rs/Rv) values, which essentially makes the `Tabulated1Dfunction` return `nan`/`inf` when trying to evaluate. These duplicate values are now pruned when creating the table.

I have not observed any side effects on the tests on my side.

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
@@ -529,15 +529,13 @@ public:
             const Evaluation& f = RsTable.eval(pSat, /*extrapolate=*/true) - Rs;
             const Evaluation& fPrime = RsTable.evalDerivative(pSat, /*extrapolate=*/true);
 
-            const Evaluation& delta = f/fPrime;
-
-            if (!Opm::isfinite(delta)) {
-                 std::stringstream errlog;
-                 errlog << "Obtain delta with non-finite value for Rs = " << Rs << " in the "
-                        << i << "th iteration during finding saturation pressure iteratively";
-                 OpmLog::problem("wetgas NaN delta value", errlog.str());
-                 OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+            // If the derivative is "zero" Newton will not converge,
+            // so simply return our initial guess.
+            if (std::abs(Toolbox::scalarValue(fPrime)) < 1.0e-30) {
+                return pSat;
             }
+
+            const Evaluation& delta = f/fPrime;
 
             pSat -= delta;
 
@@ -556,15 +554,18 @@ public:
         }
 
         std::stringstream errlog;
-        errlog << "Could find the oil saturation pressure for Rs = " << Rs;
-        OpmLog::problem("liveoil saturationpressure", errlog.str());
+        errlog << "Finding saturation pressure did not converge:"
+               << " pSat = " << pSat
+               << ", Rs = " << Rs;
+        OpmLog::problem("liveoil psat", errlog.str());
         OPM_THROW_NOLOG(NumericalProblem, errlog.str());
     }
 
 private:
     void updateSaturationPressure_(unsigned regionIdx)
     {
-        auto& gasDissolutionFac = saturatedGasDissolutionFactorTable_[regionIdx];
+        typedef std::pair<Scalar, Scalar> Pair;
+        const auto& gasDissolutionFac = saturatedGasDissolutionFactorTable_[regionIdx];
 
         // create the function representing saturation pressure depending of the mass
         // fraction in gas
@@ -579,9 +580,15 @@ private:
                                                /*temperature=*/Scalar(1e30),
                                                pSat);
 
-            std::pair<Scalar, Scalar> val(Rs, pSat);
+            Pair val(Rs, pSat);
             pSatSamplePoints.push_back(val);
         }
+
+        //Prune duplicate Rs values (can occur, and will cause problems in further interpolation)
+        auto x_coord_comparator = [](const Pair& a, const Pair& b) { return a.first == b.first; };
+        auto last = std::unique(pSatSamplePoints.begin(), pSatSamplePoints.end(), x_coord_comparator);
+        pSatSamplePoints.erase(last, pSatSamplePoints.end());
+
         saturationPressure_[regionIdx].setContainerOfTuples(pSatSamplePoints);
     }
 
--- a/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
@@ -277,7 +277,7 @@ public:
     {
         auto& invGasB = inverseGasB_[regionIdx];
 
-        auto& RvTable = saturatedOilVaporizationFactorTable_[regionIdx];
+        const auto& RvTable = saturatedOilVaporizationFactorTable_[regionIdx];
 
         Scalar T = 273.15 + 15.56; // [K]
 
@@ -557,15 +557,13 @@ public:
             const Evaluation& f = RvTable.eval(pSat, /*extrapolate=*/true) - Rv;
             const Evaluation& fPrime = RvTable.evalDerivative(pSat, /*extrapolate=*/true);
 
-            const Evaluation& delta = f/fPrime;
-
-            if (!Opm::isfinite(delta)) {
-                 std::stringstream errlog;
-                 errlog << "Obtain delta with non-finite value for Rv = " << Rv << " in the "
-                        << i << "th iteration during finding saturation pressure iteratively";
-                 OpmLog::problem("wetgas NaN delta value", errlog.str());
-                 OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+            // If the derivative is "zero" Newton will not converge,
+            // so simply return our initial guess.
+            if (std::abs(Toolbox::scalarValue(fPrime)) < 1.0e-30) {
+                return pSat;
             }
+
+            const Evaluation& delta = f/fPrime;
 
             pSat -= delta;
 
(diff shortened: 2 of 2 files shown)


atgeirr:

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -272,7 +272,12 @@ class Tabulated1DFunction
         Scalar y0 = yValues_[segIdx];
         Scalar y1 = yValues_[segIdx + 1];
 
-        return y0 + (y1 - y0)*(x - x0)/(x1 - x0);
+        // Corner case where table is poorly defined
+        // with multiple y values for a single x value
+        if (x0 == x1)
```
atgeirr: I'd prefer braces, but I don't think @andlaus requires them...

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
```diff
@@ -529,16 +529,13 @@ class LiveOilPvt
...
-
-            if (!Opm::isfinite(delta)) {
-                 std::stringstream errlog;
-                 errlog << "Obtain delta with non-finite value for Rs = " << Rs << " in the "
-                        << i << "th iteration during finding saturation pressure iteratively";
-                 OpmLog::problem("wetgas NaN delta value", errlog.str());
-                 OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+            //If the derivative is zero Newton will not converge, so simply return our best guess instead
+            if (Toolbox::scalarValue(fPrime) == 0.0) {
+                return pSat;
```
atgeirr: I think perhaps a "better" value to return would be the lowest pressure for which the derivative is zero. It should not matter much, though.

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
```diff
@@ -557,16 +557,13 @@ class WetGasPvt
...
-
-            if (!Opm::isfinite(delta)) {
-                 std::stringstream errlog;
-                 errlog << "Obtain delta with non-finite value for Rv = " << Rv << " in the "
-                        << i << "th iteration during finding saturation pressure iteratively";
-                 OpmLog::problem("wetgas NaN delta value", errlog.str());
-                 OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+            //If the derivative is zero Newton will not converge, so simply return our best guess instead
+            if (Toolbox::scalarValue(fPrime) == 0.0) {
+                return pSat;
```
atgeirr: Same applies here: could be better to report lowest possible pressure.

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
```diff
@@ -605,9 +605,15 @@ class WetGasPvt
...
 
-            std::pair<Scalar, Scalar> val(Rv, pSat);
+            Pair val(Rv, pSat);
             pSatSamplePoints.push_back(val);
         }
+
+        //Prune duplicate Rv values (can occur, and will cause problems in further interpolation)
+        auto x_coord_comparator = [](const Pair& a, const Pair& b) { return a.first == b.first; };
+        auto last = std::unique(pSatSamplePoints.begin(), pSatSamplePoints.end(), x_coord_comparator);
+        pSatSamplePoints.erase(last, pSatSamplePoints.end());
```
atgeirr: This code (and the live oil version) should be unit tested, perhaps as simply as adding extra (redundant) data to an existing test?

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -272,7 +272,12 @@ class Tabulated1DFunction
...
+            return y0;
```
atgeirr: Is this change really necessary, when the tables are pruned?

babrodtk: I don't think it will be needed, no.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
babrodtk: My initial tests with this were not very successful. I'm running a simulation now to see if it is actually better or not.

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
babrodtk: I agree, though I am a bit uncertain as to how to make that test. I've looked at test_eclblackoilpvt.cpp but it's not clear to me how to actually design such a test.

babrodtk:
I've checked the results of this PR against a large case (with PBUB/PDEW enabled) run with master. The convergence, overall well iterations, linearizations, newton iterations, and linear iterations are identical. The solver time difference is one second after one wall clock hour of simulation, so I have not seen any performance impact either.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
babrodtk: After some testing, it appears to me that the results returning the minimum pressure, using the following code, 
```
            // If the derivative is zero Newton will not converge,
            // so simply return the smallest value.
            if (Toolbox::scalarValue(fPrime) == 0.0) {
                //Find interval where derivative is zero
                const auto& table = saturationPressure_[regionIdx];
                size_t index = RvTable.findSegmentIndex(Toolbox::scalarValue(pSat));
                Scalar x0 = RvTable.xAt(index);
                Scalar x1 = RvTable.xAt(index+1);

                //Evaluate at end points and chose minimum
                const Evaluation y0 = table.eval(x0, /*extrapolate=*/true);
                const Evaluation y1 = table.eval(x1, /*extrapolate=*/true);
                if (std::abs(Toolbox::scalarValue(y0)) < std::abs(Toolbox::scalarValue(y1))) {
                    return y0;
                }
                else {
                    return y1;
                }
            }
```
gives identical results to just returning pSat. This, however, requires that the previously private function Tabulated1DFunction::findSegmentIndex(value) is made public. I'm open for suggestions on what is best here.

opm/material/common/Tabulated1DFunction.hpp
andlaus: right: I don't care, so I'll just follow the judgement of the patch's author. I'm less lenient about identation, though.(which is fine here, BTW)

opm/material/common/Tabulated1DFunction.hpp
andlaus: Because it is simpler I prefer to have it removed in this case.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
```diff
@@ -529,16 +529,13 @@ class LiveOilPvt
...
-            const Evaluation& delta = f/fPrime;
-
-            if (!Opm::isfinite(delta)) {
-                 std::stringstream errlog;
-                 errlog << "Obtain delta with non-finite value for Rs = " << Rs << " in the "
-                        << i << "th iteration during finding saturation pressure iteratively";
-                 OpmLog::problem("wetgas NaN delta value", errlog.str());
-                 OPM_THROW_NOLOG(NumericalProblem, errlog.str());
+            //If the derivative is zero Newton will not converge, so simply return our best guess instead
+            if (Toolbox::scalarValue(fPrime) == 0.0) {
```
andlaus: I suppose you should do this using

```c++
if (std::abs(Toolbox::scalarValue(fPrime)) < 1e-30)
```

otherwise GCC spits out some warnings?

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
```diff
@@ -579,9 +579,15 @@ class LiveOilPvt
...
 
-            std::pair<Scalar, Scalar> val(Rs, pSat);
+            Pair val(Rs, pSat);
             pSatSamplePoints.push_back(val);
         }
+
+        //Prune duplicate Rs values (can occur, and will cause problems in further interpolation)
+        auto x_coord_comparator = [](const Pair& a, const Pair& b) { return a.first == b.first; };
+        auto last = std::unique(pSatSamplePoints.begin(), pSatSamplePoints.end(), x_coord_comparator);
+        pSatSamplePoints.erase(last, pSatSamplePoints.end());
```
andlaus: we should think about whether this should be done by `Opm::Tabulated1DFunction` and `Opm::Spline`...

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
andlaus: doesn't GCC produce warnings here?

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
babrodtk: I don't really understand what you mean here. Do you mean that both Tabulated1DFunction and Spline should do this, or that perhaps it should be done somewhere else?

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
babrodtk: I didn't notice a warning with this one, but I can add the fix you suggest.

andlaus: yeah, some versions produce it, some don't. I guess that if you set `-Wextra`, exact float comparisons will always be complained about, though.

opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp
andlaus: I mean the former, but I'm not sure because I consider having to eliminating such duplicates to be quite a hack. (i.e., IMO the input values can be considered to be broken if the x-values are not strictly monotonic in the context of these classes.)

babrodtk: I'm not entirely sure how this will work out with the splines, since these are end points and would hence affect more than the interval with zero slope. The issue is that these tables are created based upon evaluations of other tables again. 

I think the actual table is created in ```LiveOilPvt::updateSaturationPressure_(...)```, and if the table is monotonic (but not strictly increasing/decreasing), we may end up with "duplicate" Rs values (four in the case that I was examining) for different pSat values.

andlaus: I think that it would be identical for splines, in particular for monotonic ones. that said, I agree that it is probably not a good idea.

atgeirr:
Will merge now as it addresses a problem with model 2 (OPM/opm-simulators#1120). The missing part is that I wish it had unit tests. However, given the lack of unit tests here (other than API tests) it is nontrivial to add. This is a real problem. We have however allocated effort to address this within a reasonably short timeframe.

PR #224, opened 2017-04-11
terminal output modifications

nairr:
minor changes to terminal output message tag and classification


atgeirr:
I agree with this change, and think it should be merged when tests run. There is one caveat though: for Flow, these functions are only used for output, so it makes sense to classify the messages as warnings. It would however be possible to use a different formulation (with different primary variables), in which these functions would be used as part of the residual assembly. For such cases these messages would better be classified as problems... For now I think the current PR is doing the right thing, though.

andlaus:
ok, this change is rather simple and travis is happy with it. going in.

PR #230, opened 2017-05-07
Use libecl in jenkins.

joakim-hove:


andlaus:
I suppose that this requires coordinating the merges. for ewoms and opm-material, merge whenever.

PR #246, opened 2017-09-04
Allow for 2p gasoil case

totto82:


andlaus:
this looks harmless enough. merging.

PR #247, opened 2017-09-15
Cmakify sibling search

blattms:
This PR does the following:

0. Deactivates tabs for emacs.
1. Introduces ther mandatory project() call in the top level CmakeLists.txt to make all variables properly defined.
2. If sibling search is activated and opm-common_DIR is not set, then we try to determine the build directory layout from ${PROJECT_BINARY_DIR}. The following two possibilities are supported:
   + modules-build-dir/project-name
   + project-name/build-dir

   where project-name is the case sensitive module name (in this case opm-common).
   This results in the following search precedence:
   1. User set opm-common_DIR 
   2. sibling directories search (if the directories exist)
   3. Default (system) CMake search path

  The old handling using OPM_COMMON_ROOT is still there. This can be removed in the future.
3. Added warnings if opm-common_DIR does not exists or OPM_COMMON_ROOT is used


blattms:
For the other modules similar PRs will be posted. I just started with opm-material because it has less dependencies and seemed to be a nice play/testing ground.

blattms:
This PR might be a bit restrictive as it assumes that opm-material is actually cloned into a directory with that name.  opm-material1 would not work. Same  for building in build-dir/module-name-with-typo.Maybe we should strip some if statements?

blattms:
jenkins build this please

blattms:
No assumptions about the name of the source directory of the current module are made anymore (but for opm-common) and build directories named like <prefix><module-name><postfix> are now supported.

CMakeLists.txt
```diff
@@ -40,6 +40,11 @@ if(NOT OPM_COMMON_ROOT)
         set(opm-common_DIR "${_parent_full_dir}/opm-common}")
       endif()
     endif()
+  else()
+    if(NOT IS_DIRECTORY ${opm-common_DIR})
```
akva2: amend commit message please (waring -> warning)

CMakeLists.txt
```diff
@@ -14,6 +14,9 @@
 #                                                                         #
 ###########################################################################
 
+# Mandatory call to project
+project(opm-material CXX)
```
akva2: *blush* how could i have missed this for all these years. this is definitely correct.

CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ cmake_minimum_required (VERSION 2.8)
 set( OPM_COMMON_ROOT "" CACHE PATH "Root directory containing OPM related cmake modules")
 option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 
+if(OPM_COMMON_ROOT AND opm-common_DIR)
```
akva2: i vote we kill the _ROOT usage completely. it's legacy cmake stuff. the proper way to go about it is to use CMAKE_PREFIX_PATH entries.

blattms: Fine by me. Sounds great. Will add a patch.

blattms:
All comments should be addressed (Droped OPM_COMM_ROOT, waring -> warning)

akva2:
great. i would appreciate if we could do a synced merge across repos. would allow better overview (how to do the 'generic' sibling stuff etc).

atgeirr:
Sorry if this already was made clear, but do I need to change my build structure? Currently I have two build dirs, opm-build-debug and opm-build-release. Each contains a set of softlinks to the source code of each module (opm-common, opm-material etc.) and a set of build directories named opm-common-build, opm-material-build etc.).

I am not refusing to change it, I just would like to know if I have to do it... Also, being able to build both debug and release builds based on a single source is a necessity. My softlink trick works well with the current build system, but I would not be opposed to simplifying changes.

blattms:
If I understand your use case correctly, it should work like a charm
with these changes. With the patch could even get rid of the
softlinks. I am not sure why you need them now, though. Is it
convenience for you or does the current build system really need them?

bska:
> I am not sure why you need [symbolic links to source directories], though. Is it convenience for you or does the current build system really need them?

"It depends".  If all source trees are checked out as sibling directories inside a single directory (e.g., inside `${HOME}/work/opm/src`) then those symbolic links are entirely redundant and could (and should) be removed.  The sibling search will discover the source trees of requisite upstream modules.  If the source trees are **not** checked out into a single host directory then symbolic links in the build tree may well be needed.  I have not tried such a setup.

blattms:
I have posted similar PRs for the other modules. I hope I did not miss any module. These patches are independent of each other. After these are merged I will prepare a PR for opm-common that changes and simplifies how we find opm-modules apart from opm-common.

CMakeLists.txt
```diff
@@ -14,45 +14,43 @@
...
+    endif()
+  endif()
+else()
+  if(NOT IS_DIRECTORY ${opm-common_DIR})
+    message(WARNING "Value ${opm-common_DIR} passed to variable"
+      " opm-common_DIR is not a directory")
+  endif()
 endif()
 
-if (opm-common_FOUND)
```
atgeirr: I just wondered: if opm-common cannot be found now, what error message is displayed to the user? I kinda liked the message deleted below -- of course it had to be removed since it was no longer accurate, but I liked the hand-holding going on there.

blattms:
Here you go.This is the error message you (hopefully never) get:

```
CMake Error at CMakeLists.txt:51 (find_package):
  By not providing "Findopm-common.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "opm-common", but CMake did not find one.

  Could not find a package configuration file provided by "opm-common" with
  any of the following names:

    opm-commonConfig.cmake
    opm-common-config.cmake

  Add the installation prefix of "opm-common" to CMAKE_PREFIX_PATH or set
  "opm-common_DIR" to a directory containing one of the above files.  If
  "opm-common" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
```

atgeirr:
That message looks good enough! Also, it looks like it is a standard CMake message so its meaning could be googled by an unlucky user.

I am happy with this, so @akva2 you have the button.

CMakeLists.txt
```diff
@@ -34,24 +34,29 @@ if(NOT OPM_COMMON_ROOT)
...
-      # we assume <build-dir>/<module-name>
-      if(IS_DIRECTORY ${_parent_full_dir}/opm-common)
-        set(opm-common_DIR "${_parent_full_dir}/opm-common}")
-      endif()
+    #Try if <module-name>/<build-dir> is used
+    get_filename_component(_modules_dir ${_parent_full_dir} DIRECTORY)
+    if(IS_DIRECTORY ${_modules_dir}/opm-common/${_leaf_dir_name})
+      set(opm-common_DIR ${_modules_dir}/opm-common/${_leaf_dir_name})
+		else()
+			string(REPLACE ${PROJECT_NAME} opm-common _opm_common_leaf ${_leaf_dir_name})
```
akva2: sorry for nitpick but tabs here, spaces in other repos. please settle on one (idc which).

blattms: That seems to be an outdated diff. The tabs have already been removed and I forced pushed the changes. Sorry for not mentioning this.

PR #248, opened 2017-09-21
SWATINIT Also scale capPres when water saturation is below residual

totto82:
This assure that the initial state is in eqvilibrium also for cells with
where SWATINIT is set below residual value.


totto82:

alfbr:
Thanks for the fix, I will try to get this tested tomorrow.

alfbr:
First test results are in. I am afraid this patch slightly deteriorate water production curves in select wells. Maybe we need a chat about what I should look for.

totto82:
I have updated the equil test to comply to this PR.

andlaus:
looks good, permission to merge granted

PR #249, opened 2017-09-25
update packaging to reflect new libecl

akva2:


atgeirr:
I have checked this set of PRs for silly mistakes and could not spot any. Since they only affect packaging and that is not tested by Jenkins it makes no sense to run them through that gauntlet either. According to @akva2 the PRs have been tested locally to work correctly. I'll therefore merge the set.

Thanks!

PR #251, opened 2017-10-17
add mpi to debian packaging

akva2:
reorganizations in the build system made this rear its head. we get HAVE_MPI in the config file for opm-material, and it will be off, which is inconsistent with other modules.


andlaus:
thanks.

PR #252, opened 2017-10-24
fixed: build with MPI

akva2:
opm-material exports MPI related settings in its config file,
this does not really make sense, but enable in packaging for now.


andlaus:
thanks. I guess you already pushed this to the release branch?

akva2:
yes this was a backport.

PR #257, opened 2017-11-13
disable high performance mpi transport in debian packaging

akva2:
doesn't work inside pbuilder when building for stretch. this is already
there for the other mpi-enable modules, but was forgotten
here when MPI was turned on for the tests.


atgeirr:
Please merge when green.

BTW, what are the consequences for debian packages? Does it mean that if you want to use special cluster interconnects etc. you cannot use the binaries but need to compile from scratch?

akva2:
no, it's just the execution. tests have to run through mpirun due to stupidities in certain (open)mpi versions. it might even not be necessary any longer, but let's be on the safe side. none of this touches anything on jenkins so then i'll just merge.

PR #260, opened 2017-11-23
Add support for SOF2 for 2p runs.

totto82:
Make it possible to use SOF2 in combination with either
SGFN og SWFN for gas-oil and water-oil case respectively.

Need OPM/opm-parser#1157 for the tests to pass.


andlaus:
Looks good, merge at will

andlaus:
jenkins build this opm-parser=1157 please

andlaus:
okay, OPM/opm-parser#1157 has been merged, so this can go in too. Thanks for the patch!

andlaus:
for some reason travis does not seem to like this. (in this case not because of the usual timeout but because the compareECLFiles_flow+SPE5CASE1 opm-simulators test fails: https://travis-ci.org/OPM/opm-material/builds/306670470?utm_source=email&utm_medium=notification)

totto82:
Master was broken for SPE5CASE1 for while. Should be fixed now.

andlaus:
ok, cool

PR #262, opened 2017-11-28
Fix drsdt

hnil:
Drsdt relatert.

PR #269, opened 2017-12-19
Add support for DRSDT

totto82:
This is a rebased, cleaned and squashed version of #262 

@hnil Did I miss something? 
@andlaus Can you review this?

### DIFF
--- a/opm/material/fluidsystems/BlackOilFluidSystem.hpp
+++ b/opm/material/fluidsystems/BlackOilFluidSystem.hpp
@@ -96,6 +96,15 @@ namespace FluidSystems {
  *
  * \tparam Scalar The type used for scalar floating point values
  */
+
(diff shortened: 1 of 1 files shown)


opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -211,6 +220,20 @@ class BlackOil : public BaseFluidSystem<Scalar, BlackOil<Scalar> >
         assert(numActivePhases_ >= 2 && numActivePhases_ <= 3);
 
         setEnableDissolvedGas(deck.hasKeyword("DISGAS"));
+        if(deck.hasKeyword("DRSDT")){
```
andlaus: since this is a time related quantity, this does not belong into the fluid system. (fluid systems should only be concerned with thermodynamics.)

PR #271, opened 2017-12-26
Removed travis integration

joakim-hove:


andlaus:
thanks. I'll merge this now because it does not seem to affect travis integration of any upstream or downstream modules, only that of opm-material itself.

PR #281, opened 2018-01-24
fixed: add HAVE_OPM_PARSER to config vars

akva2:
Reared its head while doing https://github.com/OPM/opm-parser/pull/1199


andlaus:
thanks.

PR #284, opened 2018-01-30
Silence some warnings

atgeirr:
Mostly unused arguments.

(Note to the diligent reviewer: yes, I accidentally used my master branch for this, which is not optimal...)


andlaus:
thanks. I wonder why the compiler did not complain about the unused arguments earlier.

GitPaean:
> thanks. I wonder why the compiler did not complain about the unused arguments earlier.

 We saw compiler did not report unused variable warning in several different places, and later spotted by a different compiler or different version of gcc.

PR #298, opened 2018-05-23
fixed: explicitly add boost::filesystem to prereqs

akva2:
required due to used headers from opm-common


akva2:
don't think this is controversial, it's buildsystem and i want to verify the integration setup for rh6 now => shameless self-merge.

PR #299, opened 2018-06-13
Adding a Evaluation type can have variable number of primary variables at running - time

GitPaean:
One of our ongoing developments requires we can decide the number of the primary variables when running. 

More specifically, we assign primary variables for each perforation, which will only be known when we run the specific case. 

`EvaluationV` is created for this purpose, we can `resize` the number of the variables related to it.

Without knowing the number of variables, we can not create a `EvauationV` variable just based on a scalar value. While I still created a constructor with a `scalar` for it due to something related to `std::is_constructible` in `MathToolbox.hpp`. 

@andlaus, you are familiar with `Evaluation` related, please provide some suggestions.


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

opm/material/common/MathToolbox.hpp
```diff
@@ -230,10 +230,20 @@ template <class Evaluation, class Scalar>
 Evaluation constant(const Scalar& value)
 { return Opm::MathToolbox<Evaluation>::createConstant(value); }
 
+template <class Evaluation, int numVars, class Scalar>
```
andlaus: `numVars` must become a runtime parameter here, or else the whole excercise would be pointless. also, this function should only be available for Evaluations where the number of derivatives is runtime specified and conversely the functions for compile time Evaluations should only be available for these, i.e., please add `std::enable_if`.

opm/material/densead/EvaluationV.hpp
```diff
@@ -0,0 +1,602 @@
...
+
+namespace Opm {
+namespace DenseAd {
+
+/*!
+ * \brief Represents a function evaluation and its derivatives w.r.t. a fixed set of
+ *        variables.
+ */
+template <class ValueT>
+class EvaluationV
```
andlaus: it is better to generate a specialization for `Evaluation` with `-1` as the number of parameters IMO. confer https://github.com/OPM/opm-material/blob/master/bin/genEvalSpecializations.py .

opm/material/densead/EvaluationV.hpp
```diff
@@ -0,0 +1,602 @@
...
+    template <class RhsValueType>
+    EvaluationV(const RhsValueType& c, const int varPos, const int numVars)
+    {
+        resize(numVars);
+
+        // The variable position must be in represented by the given variable descriptor
+        assert(0 <= varPos && varPos < size);
+
+        setValue( c );
+        // clearDerivatives();
```
andlaus: please add a comment before the call to resize() that this sets the derivatives to 0.

opm/material/densead/EvaluationV.hpp
```diff
@@ -0,0 +1,602 @@
...
+ */
+template <class ValueT>
+class EvaluationV
+{
+public:
+    //! field type
+    typedef ValueT ValueType;
+
+    //! number of derivatives
+    int size = 0;
```
andlaus: better make `size` a method, possibly renaming it to `numDerivatives()` because the current name is confusing IMO. (also for compile time evaluations.)

opm/material/densead/EvaluationV.hpp
```diff
@@ -0,0 +1,602 @@
...
+    //
+    // i.e., f(x) = c. this implies an evaluation with the given value and all
+    // derivatives being zero.
+    template <class RhsValueType>
+    EvaluationV(const RhsValueType& c, const int numVars)
+    {
+        resize(numVars);
+        setValue( c );
+        // clearDerivatives();
+        Valgrind::CheckDefined( data_ );
```
andlaus: for objects like `std::vector` which store data on the heap, the valgrind client request does not check the actual data!

opm/material/densead/EvaluationV.hpp
```diff
@@ -0,0 +1,602 @@
...
+template <class ValueType>
+std::ostream& operator<<(std::ostream& os, const EvaluationV<ValueType>& eval)
+{
+    os << eval.value();
+    return os;
+}
+} } // namespace DenseAd, Opm
+
+// In Dune 2.3, the EvaluationV.hpp header must be included before the fmatrix.hh
+// header. Dune 2.4+ does not suffer from this because of some c++-foo.
```
andlaus: hm, since dune 2.3 support bit the dust a while ago, these hacks should probably be removed.

opm/material/densead/MathV.hpp
```diff
@@ -0,0 +1,512 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
```
andlaus: if you implement my suggestion from above, this file should not be necessary. maybe the normal Math.hpp needs a few modifications, though.

tests/test_denseadv.cpp
```diff
@@ -0,0 +1,662 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
```
andlaus: almost the same thing for this file: with some small modifications, the standard `test_densad.cpp` should be usable.

(in the context of OPM/ewoms#318, I've heard that additional complexity is considered to be a bad thing. your current solution constitutes a lot of additional complexity and also unnecessarly duplicated code.)

opm/material/densead/EvaluationV.hpp
GitPaean: I am not sure about this.  What does physically mean an `Evaluation` with `-1` number of the variables?

opm/material/densead/EvaluationV.hpp
GitPaean: Yes. It should have a better name.

opm/material/densead/EvaluationV.hpp
GitPaean: Okay. I will remove all the `Valgrind::CheckDefined` related for this class.

GitPaean:
@andlaus ,  thanks for the comments. I think it is something I need some help to make it good to go in. 

`EvaluationV` is basically working on the simulation side. I need to finish the implementation (get the desired result) on the simulation side first, then I will come back to this PR.

andlaus:
if you want, I'll adapt the code generator for you. I'm also happy if you want to give it a shot yourself; as you prefer.

GitPaean:
> if you want, I'll adapt the code generator for you. I'm also happy if you want to give it a shot yourself; as you prefer.

Please go ahead. It is great if you would like to help.

opm/material/densead/EvaluationV.hpp
andlaus: > What does physically mean an Evaluation with -1 number of the variables?

That's our choice to decide. I propose "number of derivatives is determined at runtime".

andlaus: if you don't like this to be a raw number, a constant can be added, e.g.:

```c++
typedef Opm::DenseAd::Evaluation<double, Opm::DenseAd::runtimeDetermined> EvaluationV;
EvaluationV x(/*numDeriv=*/2);
```

GitPaean: > That's our choice to decide. I propose "number of derivatives is determined at runtime".

Okay. You are suggesting to use specialization of `Evaluation` with -1 number of variables instead of a new `EvaluationV`.  I do not have problem with that, as long as it will work for my purpose.

GitPaean:
@andlaus , when you have time, please come with your suggestion for the purpose of this PR, I am not sure I caught all your opinions on how to handle this and I am still working on the simulator side. 

The simulation side results look fine now, although quite of some improvements remain to be done, like acceleration of the convergence.

andlaus:
yeah, this is the next thing on my agenda. recently I was busy with OPM/opm-simulators#1512. (would be nice if you would give it a shot, BTW.)

GitPaean:
> yeah, this is the next thing on my agenda. recently I was busy with OPM/opm-simulators#1512. (would be nice if you would give it a shot, BTW.)

My problem is that I do not have the whole idea about how you want to do it. If I choose, I might stick with the current implementation and looking into improving it. I notice the problem of duplicating  code, which I agree is bad, which is why I look for suggestions. 

Or if you can begin with something, I can finish it if you prefer.

GitPaean:
closing this through #302 .

PR #309, opened 2018-10-17
Mixing energy

hnil:
Modifications to make energy module for black oil to not only gennerate termperature due to the entalpy/work term. This is physical since we assume the energy density only depend on temperature. It is introduced energy density on each phase depending on the phases.

### DIFF
--- a/opm/material/fluidstates/BlackOilFluidState.hpp
+++ b/opm/material/fluidstates/BlackOilFluidState.hpp
@@ -322,9 +322,11 @@ public:
      */
     Scalar internalEnergy(unsigned phaseIdx OPM_UNUSED) const
     {
-        //double ent_fac=0.0;
-        double ent_fac=1.0;
-        return (*enthalpy_)[phaseIdx] - ent_fac*pressure(phaseIdx)/density(phaseIdx);
+        //double entFac=0.0;
+        double entFac=1.0;
+        // this may be setting which for entFac =1 make this probably E100 compatible
+        // however this has to be consitent with call to enthalpy
+        return (*enthalpy_)[phaseIdx] - entFac*pressure(phaseIdx)/density(phaseIdx);
     }
 
     //////
--- a/opm/material/fluidsystems/BlackOilFluidSystem.hpp
+++ b/opm/material/fluidsystems/BlackOilFluidSystem.hpp
@@ -1030,13 +1030,9 @@ public:
         assert(0 <= regionIdx && regionIdx <= numRegions());
 
         const auto& p = Opm::decay<LhsEval>(fluidState.pressure(phaseIdx));
-        //const auto& T = Opm::decay<LhsEval>(fluidState.temperature(phaseIdx));
-        //double ent_fac=0;
-        double ent_fac = 1.0;
+        double entFac = 1.0;
         return energy<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx)
-                + ent_fac*p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
-
-        //return energy;
+                + entFac*p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
     }
 
     /*!


opm/material/fluidstates/BlackOilFluidState.hpp
```diff
@@ -321,7 +321,11 @@ class BlackOilFluidState
      * exception!
      */
     Scalar internalEnergy(unsigned phaseIdx OPM_UNUSED) const
-    { return (*enthalpy_)[phaseIdx] - pressure(phaseIdx)/density(phaseIdx); }
+    {
+        //double ent_fac=0.0;
+        double ent_fac=1.0;
```
andlaus: camelCase please

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -551,7 +551,7 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
             if (enableDissolvedGas()) {
                 // miscible oil
                 const LhsEval& Rs = Opm::BlackOil::template getRs_<ThisType, FluidState, LhsEval>(fluidState, regionIdx);
-                const LhsEval& bo = oilPvt_->inverseFormationVolumeFactor(regionIdx, T, p, Rs);
+                const LhsEval& bo = inverseFormationVolumeFactor(fluidState, phaseIdx, regionIdx);
```
andlaus: Isn't the new version identical to the old one? if no, the `Rs` variable above should be unused?!

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -655,6 +654,74 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
         throw std::logic_error("Unhandled phase index "+std::to_string(phaseIdx));
     }
 
+    template <class FluidState, class LhsEval = typename FluidState::Scalar>
+    static LhsEval energy(const FluidState& fluidState,
+                                    unsigned phaseIdx,
+                                    unsigned regionIdx)
+    {
```
andlaus: that function should be named either `internalEnergy()` or `enthalpy()`. Also, `internalEnergy()` and `enthalpy()` need to be consistent, i.e. h_alpha = u_alpha + p_alpha/rho_alpha.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -963,28 +1030,13 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
...
-                oilPvt_->internalEnergy(regionIdx, T, p, Opm::BlackOil::template getRs_<ThisType, FluidState, LhsEval>(fluidState, regionIdx))
-                + p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
-
-        case gasPhaseIdx:
-            return
-                gasPvt_->internalEnergy(regionIdx, T, p, Opm::BlackOil::template getRv_<ThisType, FluidState, LhsEval>(fluidState, regionIdx))
-                + p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
+        //const auto& T = Opm::decay<LhsEval>(fluidState.temperature(phaseIdx));
+        //double ent_fac=0;
+        double ent_fac = 1.0;
```
andlaus: camelCase please.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -963,28 +1030,13 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
...
-            return
-                oilPvt_->internalEnergy(regionIdx, T, p, Opm::BlackOil::template getRs_<ThisType, FluidState, LhsEval>(fluidState, regionIdx))
-                + p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
-
-        case gasPhaseIdx:
-            return
-                gasPvt_->internalEnergy(regionIdx, T, p, Opm::BlackOil::template getRv_<ThisType, FluidState, LhsEval>(fluidState, regionIdx))
-                + p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
+        //const auto& T = Opm::decay<LhsEval>(fluidState.temperature(phaseIdx));
+        //double ent_fac=0;
```
andlaus: please remove these two lines.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -963,28 +1030,13 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
...
-        case waterPhaseIdx:
-            return
-                waterPvt_->internalEnergy(regionIdx, T, p)
-                + p/density<FluidState, LhsEval>(fluidState, phaseIdx, regionIdx);
-
-        default: throw std::logic_error("Unhandled phase index "+std::to_string(phaseIdx));
-        }
-
-        throw std::logic_error("Unhandled phase index "+std::to_string(phaseIdx));
+        //return energy;
```
andlaus: remove

opm/material/fluidstates/BlackOilFluidState.hpp
hnil: in principle this is the same but internalEnergy only forward to a stored entalpy. For eclipse type of models it is probably simpler to do the oposite.

opm/material/fluidsystems/BlackOilFluidSystem.hpp
andlaus: ah, okay. maybe this can even replace the ´enthalpy()` and/or `internalEnergy()` functions directly?

andlaus:
I thought about whether this is physically a good idea and I am now even more unsure: the internal energy for wet gas and live oil should be lower (or maybe higher?) than for the same amounts of pure gas and pure oil. i.e., physically the temperature should change if the two components are mixed together...

opm/material/fluidsystems/BlackOilFluidSystem.hpp
hnil: Do not think so, since since the member function do the bookkeeping  with of finding if the fluid is saturated or not. It probably had been better to take all of this directly from the state, since this will ensure consistency with the flow problem which is essential for physical behavoir.. (Else the mass in the energy equation is not consitent with the mass int the system and tempreratur can be gennerated.)

opm/material/fluidsystems/BlackOilFluidSystem.hpp
hnil: I think for consistency with E100 and some simple investigation it had been use full to have this entfac as parameter. but it need to be consistent between the fluid state and the fluid system in the way. (If entFace = 0, enthalpy effect is ignored and one should temprature should stay within boundary conditions if I am not wrong...)

opm/material/fluidsystems/BlackOilFluidSystem.hpp
hnil: How could I take it directly form the state. (I tried but broke some tests)

opm/material/fluidsystems/BlackOilFluidSystem.hpp
andlaus: is this an ECLIPSE parameter, i.e. can it somehow be set from the deck?

andlaus:
it seems like this became stale. what's the status here? if you think we should spend time on testing this, please make it mergable again.

PR #310, opened 2018-10-18
BlackOilFluidState: Only store active phases in the two phase case

totto82:
Preliminary testings show 5% speed up for the linearization for 2p oil water case 

Needs a downstream PR to actually use this.


totto82:
jenkins build this please

andlaus:
this seems to be worth it and downstreams are unaffected if they do not explicitly enable this. merging; thanks for the patch!

PR #312, opened 2018-10-29
Pass canonical phaseIdx to setXXX() methods in fluidsystem

totto82:
The setXXX methods expects the canonicalPhaseIdx


andlaus:
oops, that's obviously a correct change. thanks for finding it, let's merge immediately!

PR #313, opened 2018-11-08
Table saturation fix ecl

hnil:
A test pull request for discussing changes in interpolation. 

If successfull: It will be possible to remove all functions related to saturatedXXXX.
It will decomple evaluation of the tables from the line of saturated values.
It test is so far to limited and the changes in the interpolation is not put into all properties.

This add 2 ways to get consistent evaluation at the saturation line:
1) interpolate according to the slop of the saturation line for oil related and according to scale slop for gas scale is 1 at the saturation line and zero at zero. This still require evaluation only inside the valid domain but I think it will always work.
2) Extend the tables by using derivatives from the line below, but interpolate as we have done all the time. This is correct for the test but may need to change the interpolation on the saturation line for some tables. The extenstion is no minimal, but do not need to be...
3) Not tested: the methods can be combind if the processing for 1) is done first. this will give equivalent results of 1 inside, but extend nicely to values slitly outside the domain.

The only on the evaluation at the saturation line so far. Serious testing need. Any good comments will be fine..


opm/material/common/UniformXTabulated2DFunction.hpp
```diff
@@ -281,11 +281,25 @@ class UniformXTabulated2DFunction
         // table ...
         unsigned i = xSegmentIndex(x, extrapolate);
         const Evaluation& alpha = xToAlpha(x, i);
-
-        unsigned j1 = ySegmentIndex(y, i, extrapolate);
-        unsigned j2 = ySegmentIndex(y, i + 1, extrapolate);
-        const Evaluation& beta1 = yToBeta(y, i, j1);
-        const Evaluation& beta2 = yToBeta(y, i + 1, j2);
+        // find upper and lower y value
+        Evaluation shift = 0.0;
```
andlaus: sorry for the stupid question, but what is the idea here?

opm/material/common/UniformXTabulated2DFunction.hpp
```diff
@@ -379,7 +393,127 @@ class UniformXTabulated2DFunction
             os << "\n";
         }
     }
+    void extendTableStart(){
```
andlaus: some comment what this method does and how it does is required before merge. Also: can it be made private?

tests/test_eclblackoilpvt_eval.cpp
```diff
@@ -0,0 +1,908 @@
...
+#include "config.h"
+
+#if !HAVE_ECL_INPUT
+#error "The test for the black oil PVT classes requires eclipse input support in opm-common"
+#endif
+
+#include <iostream>
+#include <iomanip>
+#include <cmath>
+#include <limits>
```
andlaus: the order of includes in opm-material is in decreasing order of abstractness, i.e., headers from the standard library should be included last. (this rule makes missing includes a little bit more visible.)

CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ opm_add_test(test_blackoilfluidstate)
 opm_add_test(test_ConditionalStorage)
 opm_add_test(test_eclblackoilfluidsystem CONDITION HAVE_ECL_INPUT)
 opm_add_test(test_eclblackoilpvt CONDITION HAVE_ECL_INPUT)
+opm_add_test(test_eclblackoilpvt_eval CONDITION HAVE_ECL_INPUT)
```
andlaus: BTW: this test already is conditional on the availability of the parser, so nothing needs to be done here...

hnil:
this should be substitied by #314

PR #324, opened 2019-03-06
changed fluid to put water equations first

hnil:
Move water to first equation. This make implementation of cpr simpler.

### DIFF
--- a/opm/material/fluidsystems/BlackOilFluidSystem.hpp
+++ b/opm/material/fluidsystems/BlackOilFluidSystem.hpp
@@ -402,9 +402,9 @@ public:
     static const unsigned numComponents = 3;
 
     //! Index of the oil component
-    static const unsigned oilCompIdx = 0;
+    static const unsigned oilCompIdx = 1;
     //! Index of the water component
-    static const unsigned waterCompIdx = 1;
+    static const unsigned waterCompIdx = 0;
     //! Index of the gas component
     static const unsigned gasCompIdx = 2;
 


andlaus:
making the phase indices different from the component indices was a conscious decision to catch incorrect usage. (i.e., using a phase index where a component index is required or vice versa). sounds like your CPR code currently has some issues in this respect?

GitPaean:
> making the phase indices different from the component indices was a conscious decision to catch incorrect usage.

Should we change component indices? then we will get rid of a lot indices conversion functions in `opm-simulators` .

andlaus:
> Should we change component indices? 

no. You should fix the user code. (if necessary, create a phaseToCompIdx[phaseIdx] array.)

GitPaean:
>  (if necessary, create a phaseToCompIdx[phaseIdx] array.)

That is the thing we want to avoid.

andlaus:
> Creating an array is basically the same with a conversion function.

yes, you can also create a conversion function if you prefer. phases are something very different from components, so using the same indices to access both should be considered a felony IMO.

GitPaean:
> yes, you can also create a conversion function if you prefer. phases are something very different from components, so using the same indices to access both should be considered a felony IMO.

Maybe, but the fact is the component indices are not same either.

atgeirr:
Since keeping the indices different may be a good idea to provoke errors if abused, I am testing now with changed phase indices as well. (Basically swapping oil and water for phases too.) If that works all should be happy...

andlaus:
> Maybe, but the fact is the component indices are not same either.

what do you mean with this? my point is that you should not be able to use a component index to specify a phase without doing some translation before...

@atgeirr: since the eWoms and ebos code should not mix the phase and component indices it should work fine. I'm not sure about `flow`'s well model, though.

GitPaean:
> what do you mean with this?

like the function `ebosCompIdxToFlowCompIdx()`. I am not familiar with ebos enough to comment to dare to change these functions.  We got rid of quite some already, but there are still some that makes people nerves.

atgeirr:
> @atgeirr: since the eWoms and ebos code should not mix the phase and component indices it should work fine. I'm not sure about `flow`'s well model, though.

It compiles fine, but I get plenty of regression errors. Maybe there are some abuses in the output code? Not sure, needs checking unfortunately.

GitPaean:
> what do you mean with this? my point is that you should not be able to use a component index to specify a phase without doing some translation before...

I now see the discussion went to the wrong direction due to me. I will shut up now. Sorry for the noise.

andlaus:
> Maybe there are some abuses in the output code? 

that's very possible. (Sorry, I forgot that part in my previous comment.)

andlaus:
this PR might have a point: to catch such index mixups in the current code, maybe the component indices should be changed after all. They should not be the same as the phase indices with the same name, though. Maybe to something like 0 for gas, 1 for water and 2 for oil?

andlaus:
to be clear: While I won't allow the appearance of generality for user code that essentially hard-codes the indices, I'm not objecting, to changing the default order of the indices if these indices lead to better *performance*. Any arbitrary ordering must lead to *correct* results, though. (since the latter does not seem to be the case at the moment, there's a bug somewhere. I'll probably add a test using a fluid system that uses completely different indices to opm-simulators soon in order to prevent such regressions from happening.)

atgeirr:
> since the latter does not seem to be the case at the moment

Note that this applies to the phase indices, not component indices.

andlaus:
> Note that this applies to the phase indices, not component indices.

reordering the component indices is currently required for correctness with @hnil's current version of CPR, not just for performance?

atgeirr:
I have now looked into the question of why changing phase indices failed, and it is was simply because I did not realize I also had to change the `phaseToSolventCompIdx_` and `phaseToSoluteCompIdx_` arrays. The fluid system test in opm-material ensures this, so no need to add anything. With this, it seems everything (except the equil test, which has hardcoded indices that must be fixed -- PR coming) works with changed indices, which is a good sign!

Knowing that changing phase indices (so they are different from component indices) should work fine, I'll run tests for this PR.

atgeirr:
jenkins build this please

andlaus:
this is superseeded by #326.

PR #326, opened 2019-03-11
Put water component first.

atgeirr:
This will put the water equation first in Flow. Based on #325, replacing #324.

### DIFF
--- a/opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp
+++ b/opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp
@@ -36,17 +36,17 @@ namespace Opm {
 class BlackOilDefaultIndexTraits
 {
 public:
-    //! Index of the water phase
-    static const unsigned waterPhaseIdx = 0;
     //! Index of the oil phase
-    static const unsigned oilPhaseIdx = 1;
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

andlaus:
In principle I'm okay to merge this once the the bug mentioned in OPM/opm-simulators#1754 has been fixed, i.e., `ebos` and `ebos_altidx` produce identical results (modulo numerical noise) . I'd still advice you to implement an `ebos_cpr` variant for this instead of dragging everybody along, but I'm not dogmatic about this issue as long as somebody else is willing to take the blame in case something blows up...

opm/material/fluidsystems/BlackOilDefaultIndexTraits.hpp
```diff
@@ -37,16 +37,16 @@ class BlackOilDefaultIndexTraits
 {
 public:
     //! Index of the water phase
-    static const unsigned waterPhaseIdx = 0;
+    static const unsigned oilPhaseIdx = 0;
     //! Index of the oil phase
```
andlaus: the comments need to be adapted as well

andlaus:
I guess that with OPM/opm-simulators#1757 merged, this is no longer required. Closing tentatively.

PR #336, opened 2019-05-22
Enable onephase

hnil:
WIP to make onephase with and without termal running.
For all to work it  need ewoms and opm_simulator with same branch name,  the test is outside of the current framework but just changes tags.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -448,10 +448,11 @@ private:
             + (oilEnabled?1:0)
             + (waterEnabled?1:0);
 
-        if (numEnabled < 2)
+        if(numEnabled ==0){
             throw std::runtime_error("At least two fluid phases must be enabled. (Is: "+std::to_string(numEnabled)+")");
-
-        if (numEnabled == 2) {
+        }else if   (numEnabled == 1){
+            threePhaseApproach_ = Opm::EclMultiplexerApproach::EclOnePhaseApproach;
+        } else if (numEnabled == 2) {
             threePhaseApproach_ = Opm::EclTwoPhaseApproach;
             if (!gasEnabled)
                 twoPhaseApproach_ = Opm::EclTwoPhaseOilWater;
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -448,11 +448,11 @@ private:
             + (oilEnabled?1:0)
             + (waterEnabled?1:0);
 
-        if(numEnabled ==0){
+        if( numEnabled == 0 ){
             throw std::runtime_error("At least two fluid phases must be enabled. (Is: "+std::to_string(numEnabled)+")");
         }else if   (numEnabled == 1){
             threePhaseApproach_ = Opm::EclMultiplexerApproach::EclOnePhaseApproach;
-        } else if (numEnabled == 2) {
+        } else if ( numEnabled == 2) {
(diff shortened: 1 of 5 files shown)


opm/material/thermal/EclThconrLaw.hpp
```diff
@@ -56,12 +56,15 @@ class EclThconrLaw
...
     {
         // THCONR + THCONSF approach.
-        Scalar lambdaRef = params.referenceTotalThermalConductivity();
-        Scalar alpha = params.dTotalThermalConductivity_dSg();
-
+        Scalar lambdaRef = params.referenceTotalThermalConductivity();        
         static constexpr int gasPhaseIdx = FluidSystem::gasPhaseIdx;
-        const Evaluation& Sg = Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
-        return lambdaRef*(1.0 - alpha*Sg);
+        if(FluidSystem::phaseIsActive(gasPhaseIdx)){
```
andlaus: I'm not terribly happy with this code as it hard-codes phase names and it depends on a non-standard method in the FluidSystem (the former was also the case before). it would be nice at least the fluid system API situation would not be regressed: can you leave it as-is and just not compute the thermal conductivity for inactive phases in the intensive quantities?

joakim-hove:
> and the ordering of the commits is not terribly logical as far as I can see.

I don't know if this is what happens here - but observe that github ui presents the commits in a strictly cronological fashion and does not reflect reordering of commits as part of a rebase - that is *really broken* in my opinion.

andlaus:
I don't mind to use github's squash merge functionality instead of doing the history cleaning during review. the commit message might be a bit messed up with this, though...

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -448,10 +448,11 @@ class EclMaterialLawManager
             + (oilEnabled?1:0)
             + (waterEnabled?1:0);
 
-        if (numEnabled < 2)
+        if(numEnabled ==0){
```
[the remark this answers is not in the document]
hnil: fixed

opm/material/fluidsystems/BlackOilFluidSystem.hpp
```diff
@@ -199,7 +199,7 @@ class BlackOilFluidSystem : public BaseFluidSystem<Scalar, BlackOilFluidSystem<S
         setReservoirTemperature(eclState.getTableManager().rtemp());
 
         // this fluidsystem only supports two or three phases
-        assert(numActivePhases_ >= 2 && numActivePhases_ <= 3);
+        //assert(numActivePhases_ >= 2 && numActivePhases_ <= 3);
```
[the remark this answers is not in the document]
hnil: fixed

opm/material/thermal/EclThconrLaw.hpp
hnil: For me it seems lik this is called for all thermal runs and would make problems for cases where gas i not active. I think the change is needed. What is the non standard names?

andlaus: yes, the point was that the calling code should not need this quantity in the first place, i.e. arguably there is a small bug there!?

> What is the non standard names?

everything which is blackoil specific, i.e., everything which does not make sense if you go to "base thermodynamics". such stuff is e.g., the gas dissolution factor, the concept of "active" phases or even the names of phases or components. If you want to find out if black-oil specific stuff crept into your code, you can try to compile it with e.g. [`TwoPhaseImmiscibleFluidSystem`](https://github.com/OPM/opm-material/blob/master/opm/material/fluidsystems/TwoPhaseImmiscibleFluidSystem.hpp)

atgeirr:
Closing in favour of #342, which is a revised and squished version.

PR #338, opened 2019-06-07
fill value zero for Evaluation(int numDerivatives) of DynamicEvaluation

GitPaean:
otherwise, we get some small non-zero values there.

it causes the failure of polymer case for https://github.com/OPM/opm-simulators/pull/1760

### DIFF
--- a/bin/genEvalSpecializations.py
+++ b/bin/genEvalSpecializations.py
@@ -216,7 +216,7 @@ public:
 {% if numDerivs < 0 %}\
     // create a "blank" dynamic evaluation
     explicit Evaluation(int numDerivatives)
-        : data_(1 + numDerivatives)
+        : data_(1 + numDerivatives, 0.0)
     {}
 
     // create a dynamic evaluation which represents a constant function
--- a/opm/material/densead/DynamicEvaluation.hpp
+++ b/opm/material/densead/DynamicEvaluation.hpp
@@ -115,7 +115,7 @@ public:
 
     // create a "blank" dynamic evaluation
     explicit Evaluation(int numDerivatives)
-        : data_(1 + numDerivatives)
+        : data_(1 + numDerivatives, 0.0)
     {}
 
     // create a dynamic evaluation which represents a constant function


GitPaean:
jenkins build this please

GitPaean:
```
Build step 'Execute shell' marked build as failure
c++: internal compiler error: Terminated (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-6/README.Bugs> for instructions.
CMakeFiles/ebos_altidx.dir/build.make:62: recipe for target 'CMakeFiles/ebos_altidx.dir/ebos/ebos_altidx.cc.o' failed
make[2]: *** [CMakeFiles/ebos_altidx.dir/ebos/ebos_altidx.cc.o] Error 4
make[2]: *** Deleting file 'CMakeFiles/ebos_altidx.dir/ebos/ebos_altidx.cc.o'
[WARNINGS] Skipping publisher since build result is FAILURE
Archiving artifacts
Recording test results
ERROR: Step ?Publish JUnit test result report? failed: No test report files were found. Configuration error?
```

GitPaean:
jenkins build this please

andlaus:
seems like the jenkins server has too little RAM...

andlaus:
jenkins build this please

akva2:
it is not OOM, we have about 30GB free ram.

andlaus:
I don't mind this change much, but the current behavior of master is probably better because there, dynamic dense AD objects behave like static ones and also like build-in types, e.g., one expects `d` to be undefined here:

```c++
double d;
std::cout << d << std::endl;
```

whereas if it was a `DynamicEvaluation`, it would be a `0`-constant with this patch.

tl;dr: it is IMO better to use `Opm::DynamicEvalution x(numDeriv) = 0.0` in the calling code.

GitPaean:
Or we make the following function differently in DynamicEvaluation.  

```
168     static Evaluation createBlank(const Evaluation& x)
169     { return Evaluation(x.size(), 0.0); }
```

GitPaean:
I mean, `createBlank` should be well defined. It is a problem if it returns some undefined values inside?

andlaus:
> Or we make the following function differently in DynamicEvaluation.

yes, but my point was that if you create an evaluation object without specifying anything, it should behave like a build-in type, i.e. be uninitialized. the main reason is that you might want to use the same code for build-in types and dense-AD objects by templating over said type. If you do this, you might be unpleasantly surprised once you swap out `DynamicEvaluation` for e.g. `double`. also, such errors should be easy to find with `valgrind` in the current master, but almost impossible with this PR...

GitPaean:
I mean to change this PR to be through the following change. Do you think it is a better solution?

```
168     static Evaluation createBlank(const Evaluation& x)
169     { return Evaluation(x.size(), 0); }
```

`createBlank()` function should produce well defined behavior?


I am wondering what are the intended possibilities for the template parameter `ValueT`.

andlaus:
> `createBlank()` function should produce well defined behavior?

yes, if you count "memory allocated but values undefined" as "well defined behavior". the idea of `createBlank()` is to be the equivalent of the first line of

```c++
double d;
// ...
d = 123.456; 
```

andlaus:
(if you want well-defined values, you can use `createConstant()` or `createVariable()` in addition to `auto x = createBlank(n); x = 0.0;`.)

GitPaean:
> (if you want well-defined values, you can use createConstant() or createVariable() in addition to auto x = createBlank(n); x = 0.0;.)

Sure. It is fine. It basically there are some code has been originally wrong, while it just happened not to generate problems.

andlaus:
> Sure. It is fine. It basically there are some code has been originally wrong, while it just happened not to generate problems.

I'd also wouldn't mind something like the proposed patch much, but before doing so, I'd like to be sure that it is a good idea ;)

GitPaean:
We need a solution to make the following code works for `DynamiceEvalution` in `Math.hpp`. 
```
443     static Evaluation createConstant(ValueType value)
444     { return Evaluation::createConstant(value); }
445 
446     static Evaluation createVariable(ValueType value, int varIdx)
447     { return Evaluation::createVariable(value, varIdx); }
```

andlaus:
for `DynamicEvaluation`, the only thing which you need to do is to add the size as the first argument?

andlaus:
(I mean when calling these methods)

GitPaean:
> for DynamicEvaluation, the only thing which you need to do is to add the size as the first argument?

yeah. But the code should work also for other Evaluation types, other Evaluation types do not need the size information in the interface.

andlaus:
```c++
DynamicEvalution x = createBlank(y);
x = 0.0;
```

or -- if you do not care *that* much about performance:

```c++
DynamicEvalution x = y;
x = 0.0;
```

GitPaean:
Or we add a new function is `createZeroConstant`, so from the function name, it should be something well defined. And it will be easier to use. 

I do not consider the approach in https://github.com/OPM/ewoms/pull/516 is nice.

GitPaean:
> (if you want well-defined values, you can use createConstant() or createVariable() in addition to auto x = createBlank(n); x = 0.0;.)

I do not think we have unified interface for them. DynamicEvaluation requires the number of the derivatives and others do not.

andlaus:
> I do not think we have unified interface for them. DynamicEvaluation requires the number of the derivatives and others do not.

yes, unifying the API when such a fundamental difference in the underlying assumptions exists is a challenge. IMO, we should rather try to do this anyway instead of papering over usage errors. Suggestions?

GitPaean:
> yes, unifying the API when such a fundamental difference in the underlying assumptions exists is a challenge. IMO, we should rather try to do this anyway instead of papering over usage errors. Suggestions?

This PR represents the easiest way to fix it, while it might make `createBlank()` do more than it is supposed to do. It is not a problem to me, since the initialization of C++ class is not that consistent for all the situations. 

If you do want to have a different interface to handle the situation we want a more certain way to produce a zero constant DenseAD,  either we created a function we call `createConstantZero`,  or we add a new interface `createConstant(const int numDerivatives, const ValueType value)` or createConstant(const Evaluation eval, const ValueType value)`.

atgeirr:
I think adding createConstantZero() is a good solution.

PR #339, opened 2019-06-11
adding createConstantOne and createConstantZero to Evaluations

GitPaean:
following the discussion from https://github.com/OPM/opm-material/pull/338 . 

`createConstantOne()` and `createConstantZero()` are introduced to create more well defined behavoir than `createBlank()`.

### DIFF
--- a/bin/genEvalSpecializations.py
+++ b/bin/genEvalSpecializations.py
@@ -310,6 +310,24 @@ public:
     { return Evaluation(); }
 {% endif %}\
 
+    // create an Evaluation with value and all the derivatives to be zero
+{% if numDerivs < 0 %}\
+    static Evaluation createConstantZero(const Evaluation& x)
+    { return Evaluation(x.size(), 0.0); }
+{% else %}\
+    static Evaluation createConstantZero(const Evaluation& x OPM_UNUSED)
+    { return Evaluation(0.); }
+{% endif %}\
+
+    // create an Evaluation with value to be one and all the derivatives to be zero
+{% if numDerivs < 0 %}\
+    static Evaluation createConstantOne(const Evaluation& x)
+    { return Evaluation(x.size(), 1.); }
+{% else %}\
+    static Evaluation createConstantOne(const Evaluation& x OPM_UNUSED)
+    { return Evaluation(1.); }
+{% endif %}\
+
     // create a function evaluation for a "naked" depending variable (i.e., f(x) = x)
 {% if numDerivs < 0 %}\
     template <class RhsValueType>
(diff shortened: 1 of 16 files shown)


GitPaean:
jenkins build this please

atgeirr:
I think this looks like a fair way to address the issue, leaving `createBlank` as uninitialized, it think it can be merged to quickly address the current issue. However it seems to me that "treat a DynamicEvaluation with no derivatives as a constant" is a good change that should also be implemented (making the added functions here trivial, and let them lose their arguments). In the legacy AutoDiffBlock class it was also done like that (eventually), allowing construction of a constant AD object from the constant only (and no information about the number of derivatives).

atgeirr:
@andlaus Unless you are about to implement the "treat a DynamicEvaluation with no derivatives as a constant" feature I propose to merge this tomorrow. Give me a heads up if I should rather hold.

andlaus:
> @andlaus  Unless you are about to implement the "treat a DynamicEvaluation with  no derivatives as a constant" feature I propose to merge this tomorrow.  Give me a heads up if I should rather hold

hold on for another day, I want to give a more general solution a shot. for the dynamic-evaluations-without-derivatives-as-constants idea, I started implementing it, but it requires to modify each and every low-level function/operator, so I finally agree with my past self that there's too much wriggle room for mistakes...

GitPaean:
> hold on for another day, I want to give a more general solution a shot. for the dynamic-evaluations-without-derivatives-as-constants idea, I started implementing it, but it requires to modify each and every low-level function/operator, so I finally agree with my past self that there's too much wriggle room for mistakes...

It is okay. It is good you want to try.  I consider this PR is a temporary fix.  I was thinking to try this morning, while you are the one most familiar with these code. 

The major motivation is to get the PR https://github.com/OPM/opm-simulators/pull/1760 in.  By fixing the DenseAD side, we would be able to produce the exact results as before, so we will have much smaller burden in term of testing.

atgeirr:
> I want to give a more general solution a shot

What would that be? I think merging OPM/opm-simulators#1760 is very important so if we can merge this fix today that would be good, rather than waiting for the perfect solution...

andlaus:
> What would that be?

that would be #340: call `Opm::constant<Eval>(x, 0.0)` instead of `createConstantZero(x)`. same for any other constant.

andlaus:
As far as I can see this is fully superseded by #340. closing to avoid confusion.

atgeirr:
jenkins build this please

atgeirr:
Jenkins is green, I'll merge this now. @GitPaean Use these functions to fix the acute problem but do not refactor to use them elsewhere, since it may be superseded by next week...

PR #342, opened 2019-06-18
Enable onephase

atgeirr:
This builds on and replaces #336. Essentially it takes into account what my review comments would have been.

The changes I have made compared to #336 are:
 - Fix whitespace nitpicks.
 - Improve an exception message.
 - Add switch cases for EclOnePhaseApproach to silence compiler warnings, thereby making explicit the "for one-phase, do nothing" cases.

For the thermalConductivity() function I think the proposed fix is fine, it calls no methods not already called and is a simple improvement, so I think any larger refactoring of that is out of scope for this.

With that, I think this should be ready to merge when green.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -448,10 +448,11 @@ private:
             + (oilEnabled?1:0)
             + (waterEnabled?1:0);
 
-        if (numEnabled < 2)
-            throw std::runtime_error("At least two fluid phases must be enabled. (Is: "+std::to_string(numEnabled)+")");
-
-        if (numEnabled == 2) {
+        if (numEnabled == 0) {
+            throw std::runtime_error("At least one fluid phase must be enabled. (Is: "+std::to_string(numEnabled)+")");
+        } else if (numEnabled == 1) {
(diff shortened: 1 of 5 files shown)


atgeirr:
jenkins build this please

andlaus:
I'm still not sure if the call to the non-standard `FluidSystem::phaseIsActive(gasPhaseIdx)` method is necessary in `EclThconrLaw`, because the heat conductivity out not be called for inactive phases, but since that heat conduction law is very unlikely to be used for something other than `ebos` or one of its derivates, I'll merge this anyway now. (an additional explaination why it is required for onephase but does not seem to have caused trouble for twophase would still be appreciated, though...)

andlaus:
@atgeirr: Maybe I was too trigger happy once again: I get a lot of compiler warnings akin

```
/home/and/src/opm-material/build-cmake/fake-src/opm/material/fluidmatrixinteractions/EclMultiplexerMaterial.hpp:483:9: warning: enumeration value ‘EclOnePhaseApproach’ not handled in switch [-Wswitch]
```

do you want to fix or shall I revert the patch?

atgeirr:
> I get a lot of compiler warnings

I fixed those on my system, but I'll double check. I don't think we need to revert, we'll just add more extra switch cases as necessary. I do get lots of warnings from the Evaluation headers now though about unused parameter. Guess that needs fixing, too.

akva2:
see https://github.com/OPM/opm-material/pull/343

PR #347, opened 2019-06-28
added: OPM_UNUSED_NOMPI

akva2:
for suppressing unused variable warnings when a variable
is unused if compiled without MPI

### DIFF
--- a/opm/material/common/Unused.hpp
+++ b/opm/material/common/Unused.hpp
@@ -40,4 +40,10 @@
 #define OPM_OPTIM_UNUSED
 #endif
 
+#ifdef HAVE_MPI
+#define OPM_UNUSED_NOMPI
+#else
+#define OPM_UNUSED_NOMPI OPM_UNUSED
+#endif
+
 #endif


opm/material/common/Unused.hpp
```diff
@@ -40,4 +40,10 @@
 #define OPM_OPTIM_UNUSED
 #endif
 
+#ifdef HAVE_MPI
+#define OPM_UNUSED_MPI
+#else
+#define OPM_UNUSED_MPI OPM_UNUSED
```
atgeirr: I read the macro name as "unused if we have MPI" but the meaning is "unused if we do NOT have MPI". This is opposite of for example `OPM_DEBUG_UNUSED` which means "unused in debug mode".

I therefore suggest `OPM_UNUSED_NOMPI` for this. (If we ever need to add the counterpart it would be `OPM_UNUSED_MPI`.)

akva2: i fgured we'll never ever need the opposite macro, and thus it's just a longer thing to type. but okay, will rename in the name of consistency.

andlaus:
Thanks

akva2:
why is this file in opm-material? reason i ask is that there are a few similar error triggered in opm-grid, and it feels weird to pull in opm-material there for this macro only.

andlaus:
the main reason is that opm-common is not the common code but the ECL I/O libraries and non-ECL code should not be forced to go through the pain of having to build it and its dependencies. (Note that in order to do so, the dune based build system must currently be used...)

akva2:
aha. so iow it's just politics and no real reason.

i find it immensely funny that they guy spearheading the build-absolutely-everything 30 time, one for each app, and you have to repeat this for the smallest of changes, suddenly has an issue with the 15 secs it takes to build the ecl code once :D

andlaus:
yes, mainly politics. if this stays limited to this file, I suppose that a copy could also be shipped by opm-common. keeping them synchronized feels is a bit like the entrance to maintainance hell, but if it stays limited to this simple file, it is IMO better than introducing ackward dependencies. the technically best solution would obviously be to split opm-common into the real common code and the I/O libs and move all infrastructure code which currently needs to stay in other modules to the former, but this proposal did not fly the last time we had this discussion...

andlaus:
> I find it immensely funny that they guy spearheading the build-absolutely-everything 30 time, one for each app, and you have to repeat this for the smallest of changes, suddenly has an issue with the 15 secs it takes to build the ecl code once :D

About half of the approximately 1:45min required to rebuild everything OPM without tests and applications is spend for opm-common on my (32 core) machine, and -- more importantly -- I constantly struggle with failed builds due to incompatible/missing libecl and boost libraries as well as enigmatic errors from the default OPM build system (all those things are not required for the non-ECL stuff). I can fix such issues myself, but I can assure you that I've spend many-a-day to get the stuff build on my colleagues' rigs. I strongly suspect that "casual" users will simply give up under such circumstances.

PR #356, opened 2019-10-29
Added entries to changelog concerning 2019.10 release

blattms:
Should be uncontroversial. Will merge tonight if nobody else does.


blattms:
backported to release via #357

PR #365, opened 2019-11-18
Use active vectors

joakim-hove:
This is on top of #363. 

Store only the active cell elements in the `EclEpsGridProperties`. This is part of the 3D property refactor effot. The aiim is to reduce memory usage - but for this PR there will be a (temporary ..) increase in memory usage.

PR #363, opened 2019-11-18
Extract ecl eps properties

joakim-hove:
Minor simplification for 3D refactor. The purpose of this is to safer/simpler be able to change the underlying 3D properties

PR #368, opened 2019-11-27
Extract creation of EclEpsGridProperties out of loop

joakim-hove:
Creation of `EclEpsGridProperties` object was done for every cell in the case of hysterisis. Hard work!

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -542,9 +542,11 @@ private:
                                       epsGridProperties,
                                       elemIdx);
 
-            if (enableHysteresis()) {
-                EclEpsGridProperties epsImbGridProperties(eclState, true, compressedToCartesianElemIdx);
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

bska:
Okay, but is it safe to construct a set of imbibition curve end-points if the model does not enable hysteresis?

joakim-hove:
> Okay, but is it safe to construct a set of imbibition curve end-points if the model does not enable hysteresis?

In general I guess the answer might be no; but here the `EclEpsGridProperties` object can be safely constructed irrespectively - although it will probably not be usable. 

The current PR actually just reverts this detail to the way it was before: #363 - so there is nothing new here. The offending change was in #363 - but it was with #365 it became expensive.

akva2:
i don't think that's correct, this caused additional entries to appear in init files..

joakim-hove:
> this caused additional entries to appear in init files..

I know - update pushed ;-)

joakim-hove:
jenkins build this please

PR #370, opened 2019-11-28
Fp double

joakim-hove:
Enable use of new 3D properties for the Thermal properties


joakim-hove:
jenkins build this please

PR #375, opened 2019-12-13
Use temporary global properties when accesing FieldPropsManager

joakim-hove:
This comment:

There are essentially three levels of cells:

    1. The global cartesian set of cells.
    2. The set of cells which are active in the simulation.
    3. The set of cells which are active on one particular MPI process.

The original 3D implementation worked at level 1, whereas the new implementation works at level 2. In both opm-material and opm-simulators there are numerous mappings 1 → 3, at this point I have misunderstood and thought that the properties from the new container were immediately ready for use in the simulator. That is unfortunately not the case. The datastructure used in the new container should be quite well suited for the 2 → 3 conversion, but for now I have opted to go through the global vectors and then reuse the existing 1 → 3 code.

Sums up what this PR is all about.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -126,15 +126,28 @@ public:
         // copy the SATNUM grid property. in some cases this is not necessary, but it
         // should not require much memory anyway...
 #ifdef ENABLE_3DPROPS_TESTING
-        const auto& fp = eclState.fieldProps();
-        satnumRegionArray_ = fp.get<int>("SATNUM");
-        std::transform(satnumRegionArray_.begin(), satnumRegionArray_.end(), satnumRegionArray_.begin(), [](const int& satnum_value) { return satnum_value - 1; });
+        size_t numCompressedElems = compressedToCartesianElemIdx.size();
+        satnumRegionArray_.resize(numCompressedElems);
+        if (eclState.fieldProps().has<int>("SATNUM")) {
+            const auto& satnumRawData = eclState.fieldProps().get_global<int>("SATNUM");
+            for (unsigned elemIdx = 0; elemIdx < numCompressedElems; ++elemIdx) {
+                unsigned cartesianElemIdx = static_cast<unsigned>(compressedToCartesianElemIdx[elemIdx]);
+                satnumRegionArray_[elemIdx] = satnumRawData[cartesianElemIdx] - 1;
+            }
+        }
+        else
+            std::fill(satnumRegionArray_.begin(), satnumRegionArray_.end(), 0);
 
-        if (fp.has<double>("IMBNUM")) {
-            imbnumRegionArray_ = fp.get<int>("IMBNUM");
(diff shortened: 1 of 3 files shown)


joakim-hove:
jenkins build this please

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -126,15 +126,28 @@ class EclMaterialLawManager
...
-
-        if (fp.has<double>("IMBNUM")) {
-            imbnumRegionArray_ = fp.get<int>("IMBNUM");
-            std::transform(imbnumRegionArray_.begin(), imbnumRegionArray_.end(), imbnumRegionArray_.begin(), [](const int& imbnum_value) { return imbnum_value - 1; });
-        } else
-            imbnumRegionArray_ = satnumRegionArray_;
+        size_t numCompressedElems = compressedToCartesianElemIdx.size();
+        satnumRegionArray_.resize(numCompressedElems);
+        if (eclState.fieldProps().has<int>("SATNUM")) {
+            const auto& satnumRawData = eclState.fieldProps().get_global<int>("SATNUM");
```
blattms: ???? The logic seems to changed here if there is no "SATNUM".

Why not stick with copying and transforming?

joakim-hove: I also agree that the previous was cleaner - the problem was that it was wrong ;-( So - now I have reverted to code which is structurally identical to what it was before the whole 3D refactor effort started.

There are essentially three levels of cells:

1. The global cartesian set of cells.
2. The set of cells which are active in the simulation.
3. The set of cells which are active on one particular MPI process.

The original 3D implementation worked at level 1, whereas the new implementation works at level 2. In both opm-material and opm-simulators there are numerous mappings 1 &rightarrow; 3, at this point I have misunderstood and thought that the properties from the new container were immediately ready for use in the simulator. That is unfortunately not the case. The datastructure used in the new container should be quite well suited for the 2 &rightarrow 3 conversion, but for now I have opted to go through the global vectors and then reuse the existing 1 &rightarrow; 3 code.   

Current situation is that things works 100%  in serial - and there are some differences in parallell ;-(

blattms: Indeed I did not recognize the mapping cartesian->compressed. Maybe because I did not expect a parallel fix with the subject "More hack".

With the possible logic change I meant that in the case that `eclState.fieldProps().has<int>("SATNUM")` is false the values will be zero. Not sure what they are for the other branch, though.

joakim-hove:
jenkins build this please

blattms:
One more question about the design: Why return unique_ptr in try_get `(which is not CamelCase BTW)?

IMHO it seems to make no sense to add this level on direction. Just return an empty vector and check whether it its empty in sat_func:
```
    const double * satfunc(const std::vector<double>& data, std::size_t active_index) const {
        if (data.empty)
           return nullptr
       return data->operator[](active_index);  
    }
```

joakim-hove:
> Just return an empty vector and check whether it its empty in sat_func:

yes - I agree that would be better. The whole thing is supposed to be reasonably short-lived anyway; but I will make a fix in that direction.

PR #376, opened 2019-12-16
Thermal cond

joakim-hove:
The thermal properties were previously default initialized with zero - that did not make much sense, and has been removed from the `FieldPropsManager`- so now the slightly nonsensical initialization has moved down here....


joakim-hove:
jenkins build this please

PR #380, opened 2020-01-02
Remove #ifdef ENABLE_3DPROPS_TESTING - only support FieldProps

joakim-hove:
https://github.com/OPM/opm-common/pull/1298
https://github.com/OPM/opm-simulators/pull/2213

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
@@ -201,7 +201,6 @@ public:
         }
 
 
-#ifdef ENABLE_3DPROPS_TESTING
         const auto& fp = eclState.fieldProps();
         // check if we are supposed to scale the Y axis of the capillary pressure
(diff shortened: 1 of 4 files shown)


blattms:
Crashes because somehow `ECLState::fieldProps()` returns a null pointer:
```
   at /home/mblatt/src/dune/opm/opm-material/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp:130
130	        if (eclState.fieldProps().has<int>("SATNUM")) {
(gdb) p eclState.fieldPropos()
Couldn't find method Opm::EclipseState::fieldPropos
(gdb) p eclState.fieldProps()
$3 = (const Opm::FieldPropsManager &) @0x55555a93d7d8: {
  fp = std::shared_ptr<Opm::FieldProps> (empty) = {get() = 0x0}}
```

joakim-hove:
Thanks a lot Markus! I don't have time now - but I will look into it tomorrow.

blattms:
the problem seems to be incomplete rebasing and not this PR.

PR #385, opened 2020-01-20
WIP: Use local actnum

joakim-hove:
See https://github.com/OPM/opm-simulators/pull/2289

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsGridProperties.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsGridProperties.hpp
@@ -60,12 +60,12 @@ namespace Opm {
 
 namespace {
 template <typename T>
-std::vector<T> compressed_copy(const std::vector<T>& global_vector, const std::vector<int>& compressedToCartesianElemIdx) {
+std::vector<T> compressed_copy(const std::vector<T>& input_vector, const std::vector<int>& indexmap, const std::vector<int>& compressedToCartesianElemIdx) {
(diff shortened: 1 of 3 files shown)


blattms:
I think there needs to be some general discussion about this with at least some of the developers before continuing this effort. We should have a hangout with at least @dr-robertk, some people from SINTEF and me.

As I see it (and have said in other PRs) we are currently making the code  more and more unusable with grids other than CpGrid and thus burning bridges. If my memory doesn't fail me, there has been interest from other projects (dumux?) in using or contributing to the material law manager.

@joakim-hove How is this supposed to work? We read the eclipse file, set up the computational grid,
loadbalance and then compress the properties?

What we need is:
1. Not 100% sure, but it might be the case that at least one process needs to hold the properties for the unloadbalanced grid (e.g. for IO)
2. If we look at the active cells of the EclipseGrid and our loadbalanced (and possibly refined) computational grid. Then a cell of the EclipseGrid might been seen like a region of the computational
grid consisting of one or more cells of the computational grid. (A grid might need n>=1 cells to cover an active cell of the EclipseGrid)

2) was previously captured by the mapping from cartesian to compressed cell index, but is lost with your changes. We would now need an active-to-real-cellindex mapping for all grids other than (the current, unrefined) CpGrid.

blattms:
BTW: I would love to see more descriptions.

joakim-hove:
> I think there needs to be some general discussion about this with at least some of the developers before continuing this effort. We should have a hangout with at least @dr-robertk, some people from SINTEF and me.

I have updated the PR description [here](https://github.com/OPM/opm-simulators/pull/2289)

PR #386, opened 2020-01-21
fixed: enum member in EclMultiplexerApproach clashes with EclTwoPhaseApproach enum

akva2:
fix by converting both to enum class

sigh, if there is a mine, you can be certain i will be the one to trip it..


akva2:
jenkins build this please

akva2:
this should be a trivial thing to approve.

joakim-hove:
> this should be a trivial thing to approve.

problem is that no-one feels responsible in this part of town. 

But as a trigger happy bastard I'll merge.

PR #387, opened 2020-01-22
MPI serialization support for MaterialLawManager

akva2:


akva2:
jenkins build this opm-simulators=2295 please

akva2:
mergeable?

joakim-hove:
> mergeable?

I guess so; I can only repeat my comment from yesterday: *Who is responsible is not really clear in this part of the codebase*. In my opinion it would be advantageous to split this repo in the between opm-common and opm-models/opm-simulators. But that is for another day.

That said I have one question: I had thought/hoped the serialize details would be confined to opm-simulators?

akva2:
it's not doable, i have to serialize everything that use data from the deck. that being said, it's template functions so they will not be instanced except from opm-simulators where the serializer code is available.

PR #388, opened 2020-01-24
Don't Index Out of Bounds

bska:
The `phaseIsActive_` array has size `numPhases`.  We must not form the expression `phaseIsActive_[numPhases]`.  Switch to using `fill_n()` in place of `fill()` lest we have to write
```C++
std::fill(&phaseIsActive_[0], &phaseIsActive_[0] + numPhases, v)
```


bska:
jenkins build this please

blattms:
Took me some time to get the original version. I guess the author meant to say `std::fill(&phaseIsActive_, &phaseIsActive_+numPhases,...)`

PR #390, opened 2020-01-28
Thermal fixes

akva2:
Fix some issues introduced by brine and prepare for serialization of thermal law manager.

### DIFF
--- a/opm/material/common/Tabulated1DFunction.hpp
+++ b/opm/material/common/Tabulated1DFunction.hpp
@@ -129,16 +129,17 @@ public:
                          bool sortInputs = true)
     {
         assert(x.size() == y.size());
-        assert(x.size() > 1);
 
         resizeArrays_(x.size());
-        std::copy(x.begin(), x.end(), xValues_.begin());
-        std::copy(y.begin(), y.end(), yValues_.begin());
+        if (x.size() > 0) {
+            std::copy(x.begin(), x.end(), xValues_.begin());
+            std::copy(y.begin(), y.end(), yValues_.begin());
 
-        if (sortInputs)
-            sortInput_();
-        else if (xValues_[0] > xValues_[numSamples() - 1])
-            reverseSamplingPoints_();
+            if (sortInputs)
+                sortInput_();
+            else if (xValues_[0] > xValues_[numSamples() - 1])
+                reverseSamplingPoints_();
+        }
     }
 
     /*!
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this opm-simulators=2305 please

akva2:
jenkins build this opm-simulators=2305 please

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -129,16 +129,17 @@ class Tabulated1DFunction
...
-            sortInput_();
-        else if (xValues_[0] > xValues_[numSamples() - 1])
-            reverseSamplingPoints_();
+        if (x.size() > 0) {
+            std::copy(x.begin(), x.end(), xValues_.begin());
+            std::copy(y.begin(), y.end(), yValues_.begin());
+
+            if (sortInputs)
+                sortInput_();
+            else if (xValues_[0] > xValues_[numSamples() - 1])
```
bska: What's the value of `numSamples()` here?  If it's less than `1` this will be very bad.

akva2: i haven't messed with that, but we do not hit this code (it's surrounded by a newly added if()).

akva2: (value will be 0 which is why i had to if it out). the point here is that these sanity checks are not in place when default-constructing the function, but bit now when i'm recreating the empty function in deserialization.

opm/material/thermal/EclThermalLawManager.hpp
```diff
@@ -69,29 +69,33 @@ class EclThermalLawManager
         thermalConductivityApproach_ = ThermalConductionLawParams::undefinedApproach;
     }
 
-    void initFromDeck(const Opm::Deck& deck,
-                      const Opm::EclipseState& eclState,
-                      const std::vector<int>& compressedToCartesianElemIdx)
+    void initFromDeck(const Opm::Deck& deck)
+    {
+        hasSpecRock = deck.hasKeyword("SPECROCK");
```
bska: I **think** it's possible to answer the same question from `EclipseState` too.  Something along the lines of
```C++
hasSpecRock = eclState.getTableManager().hasTables("SPECROCK");
```
We'd need to check that, obviously, but it might be something to keep in mind if we really want to reduce the occurrence of direct access to the `Deck` class.

akva2: ultimately i think the simulators should not have access to the deck, ie, we can discard it after setting up the eclipsestate, schedule and field props. i'll try this one for sure.

akva2: does the job so this PR and downstream is slightly altered now.

akva2:
jenkins build this opm-simulators=2305 please

akva2:
anything else holding this back?

bska:
Both this and the downstream look good to me.  As the tests pass I'll merge this into master.  Thanks a lot.

PR #391, opened 2020-01-29
changed: avoid allocating long vectors of zeros

akva2:


akva2:
approved and independent of the sibling pr's. merging.

PR #395, opened 2020-02-07
Use data from table manager instead of parsing Deck in Pvt setup

akva2:
More deck-in-simulator removal


akva2:
jenkins build this please

PR #396, opened 2020-02-10
remove usage of CXX11Features module

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1456


akva2:
temporary close to try to beat github into submission.

PR #397, opened 2020-02-10
fixed: use std::numeric_limits<Scalar> in one more spot

akva2:
This triggered when moving to c++17.


akva2:
jenkins build this no downstreams please

PR #404, opened 2020-02-24
Use dent

joakim-hove:
Reduce deck usage

Upstream: https://github.com/OPM/opm-common/pull/1507
Dwonstream: https://github.com/OPM/opm-simulators/pull/2362


joakim-hove:
jenkins build this opm-common=1507 please

PR #405, opened 2020-02-25
fixed: wrong indices setting up HeatCr

akva2:
Need to go through the cartesian index mapper.


akva2:
jenkins build this please

bska:
This looks good in isolation, but is there anything that prevents us from using
```C++
const std::vector<double>& heatcrData  = fp.get<double>("HEATCR");
const std::vector<double>& heatcrtData = fp.get<double>("HEATCRT");
```
and then the original assignment Expressions
```C++
// ...
heatcrElemParams.setReferenceRockHeatCapacity(heatcrData[elemIdx]);
heatcrElemParams.setDRockHeatCapacity_dT(heatcrtData[elemIdx]);
```
instead?

akva2:
well yes, as is that would break in parallel. the order here is;
1) i want to fix the bug first. 
2) then i need my parallel PR landed. 
3) then we can do what you suggest, approximately.

PR #407, opened 2020-02-26
changed: lessen deck usage in pvt setup

akva2:


akva2:
jenkins build this please

opm/material/fluidsystems/blackoilpvt/OilPvtMultiplexer.hpp
```diff
@@ -134,17 +134,17 @@ class OilPvtMultiplexer
      */
     void initFromDeck(const Deck& deck, const EclipseState& eclState)
     {
-        bool enableOil = deck.hasKeyword("OIL");
-        if (!enableOil)
+        const auto& phases = eclState.runspec().phases();
+        if (!phases.active(Phase::OIL))
```
bska: You could arguably write this as
```C++
if (!eclState.runspec().phases().active(Phase::Oil))
    return;
```
to avoid having to introduce the `phases` object into the surrounding scope.  As far as I can tell, `phases` is unused elsewhere in the function.

I have no personal preference here, so I'll leave the final call to you.  You **should** however explicitly
```C++
#include <opm/parser/eclipse/EclipseState/Runspec.hpp>
```
to ensure we have a definition of the [`Opm::Phase`](https://github.com/OPM/opm-common/blob/8233ae200cb74f72347feaa8a78835d26e4de72a/opm/parser/eclipse/EclipseState/Runspec.hpp#L34) type in scope.  The current set of headers here relies too greatly on transitive includes in my opinion.

Similarly for the `WaterPvtMultiplexer.hpp` file.

akva2: both points adressed.

akva2:
jenkins build this please

opm/material/fluidsystems/blackoilpvt/OilPvtMultiplexer.hpp
bska: > both points adressed.

👍 

Much appreciated!

bska:
Build check successful.  I'll merge into master.  Thanks a lot.

PR #408, opened 2020-02-28
Use compressed field properties

akva2:
Avoid using properties in global cartesian indices.

A by-product of this is that we use the partitioned properties in parallel.


akva2:
jenkins build this opm-simulators=2372 please

akva2:
jenkins build this opm-simulators=2372 please

akva2:
any objections?

PR #410, opened 2020-03-03
use surface conditions from InitConfig

akva2:
Downstream of https://github.com/OPM/opm-common/pull/1528


akva2:
jenkins build this opm-common=1528 opm-simulators=2393 please

PR #417, opened 2020-04-06
Grid Properties: Include FieldPropsManager

bska:
We use the manager internally (for instance in the `try_get()` helper function) and therefore need a declaration in scope.


bska:
jenkins build this please

PR #421, opened 2020-06-11
Add total saturation and capillary pressure to blackoil fluid state.

atgeirr:
This is intended for use with sequential formulations. It should cause no changes whatsoever for existing solutions, so it should be possible to merge independent of downstream changes.

### DIFF
--- a/opm/material/fluidstates/BlackOilFluidState.hpp
+++ b/opm/material/fluidstates/BlackOilFluidState.hpp
@@ -198,6 +198,20 @@ public:
     { saturation_[canonicalToStoragePhaseIndex_(phaseIdx)] = S; }
 
     /*!
+     * \brief Set the capillary pressure of a fluid phase [-].
+     */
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

atgeirr:
I can confirm that this PR makes no difference at all to the solution on a Norne run, as it should. I intend to self-merge when green. (The code is from Halvor, so not sure it is proper to call it a self-merge as such.)

totto82:
@atgeirr Did you check the performance impact for this? In general we don't want to store unnecessary data in the fluidstate. Making storage of these properties conditional will solve the possible issue.

atgeirr:
I did not see any performance impact (down to the usual not-so-great accuracy of workstation runs). Is there any reason to believe there is?

totto82:
It is probably within noise, but in general I though it wise to make the intensiveQuentities as small as possible to reduce caching.

PR #422, opened 2020-06-25
Allow for extrapolation of WATVISC curve

goncalvesmachadoc:
Fix Exception: Tried to evaluate a tabulated function outside of its range


atgeirr:
jenkins build this please

PR #423, opened 2020-07-02
make the water PVT salt dependent

totto82:
The PVTWSALT was implemented but the salt concentration was never passed to the PVT model. I have removed the interface that allows for not passing the salt concentration to assure the salt concentration is always passed to the model. This should only affect the brine specialization.

### DIFF
--- a/opm/material/fluidstates/BlackOilFluidState.hpp
+++ b/opm/material/fluidstates/BlackOilFluidState.hpp
@@ -76,6 +76,18 @@ LhsEval getInvB_(typename std::enable_if<!HasMember_invB<FluidState>::value,
         /FluidSystem::referenceDensity(phaseIdx, pvtRegionIdx);
 }
 
+OPM_GENERATE_HAS_MEMBER(saltConcentration, ) // Creates 'HasMember_saltConcentration<T>'.
+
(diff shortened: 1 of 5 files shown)


opm/material/fluidstates/BlackOilFluidState.hpp
```diff
@@ -76,6 +76,18 @@ LhsEval getInvB_(typename std::enable_if<!HasMember_invB<FluidState>::value,
         /FluidSystem::referenceDensity(phaseIdx, pvtRegionIdx);
 }
 
+OPM_GENERATE_HAS_MEMBER(saltConcentration, ) // Creates 'HasMember_saltConcentration<T>'.
+
+template <class FluidState>
+unsigned getSaltConcentration_(typename std::enable_if<HasMember_saltConcentration<FluidState>::value,
```
bska, line 82: Does salt concentration only come in discrete amounts?  If not, I think the return type should be `double` rather than `unsigned` (`int`).

totto82: Thanks for noticing.

bska:
Thanks for the updates.  I'll merge this now as it looks good to me.

PR #427, opened 2020-09-16
Add a co2brine pvt model

totto82:
The co2brine model is activated by setting CO2STOR in the RUNSPEC section (Depends on OPM/opm-common#1923)
The co2 and brine pvt properties are then computed based on relations as specified
in the CO2Brine fluidsystem.

At the current stage the oil-phase is used to model the brine. If a proper gas-water
simulator is made, the Brine PVT should be moved to the water phase.

Known issues:
- Currently the viscosity of the Brine does not depend on the composition [fixed]
- Salinity is assumed to be constant = 0.1. This value is hard-coded in the co2table.inc
For other salinities the co2table.inc needs to be re-made using Span-Wagner. [fixed]
- The Opm::pow functions call takes a large portion of the run-time. [fixed]

This is ready to be merged together with OPM/opm-common#1957


opm/material/components/co2tables.inc
```diff
@@ -1,3 +1,6 @@
+#ifndef CO2TABLES_INC
```
hnil, line 1: Maybe it would be good if the numbers below was taken from a pure inputfile or at least as an option from an inputfile.

opm/material/components/co2tables.inc
```diff
@@ -40864,13 +40867,13 @@ struct CO2Tables {
    static const double brineSalinity;
 };
 
-TabulatedFunction CO2Tables::tabulatedEnthalpy;
-TabulatedFunction CO2Tables::tabulatedDensity;
-const double CO2Tables::brineSalinity = 1.000000000000000e-01;
+inline TabulatedFunction CO2Tables::tabulatedEnthalpy;
+inline TabulatedFunction CO2Tables::tabulatedDensity;
+inline const double CO2Tables::brineSalinity = 1.000000000000000e-01;
```
hnil, line 40872: Is it possible to make this imput in a simpe way?

hnil (commented):
This seem fine. But if other input possibilities were possible it could have been made more flexible, by getting values from other sources than the XXX.DATA

opm/material/components/co2tables.inc
totto82: Currently this is a auto-generated file based on an old Span-Wagner code. We could look into if we could change the output to something more readable.

opm/material/components/co2tables.inc
totto82: The CO2 table should cover the range of interest, so I don't see the point of making it an input file. On the contrary the point of the CO2STORE option is to give the user an option of not setting up the PVT input tables themselves. If the user what to use its own table he can always use the PVTO etc. options.

totto82:
I have done some cleanup based on @osae comments.

totto82:
jenkins build this please

opm/material/components/SimpleHuDuanH2O.hpp
```diff
@@ -0,0 +1,399 @@
...
+namespace Opm {
+
+ /*!
+ * \ingroup Components
+ *
+ * \brief A simple version of  pure water with density from Hu et al. 
+ *
+ * Compared to the water formulation of IAPWS'97, this class provides
+ * a much simpler component that represents the thermodynamic
+ * properties of of pure water. This implies that the likelyhood for
```
akva2, line 51: of of

opm/material/components/SimpleHuDuanH2O.hpp
```diff
@@ -0,0 +1,399 @@
...
+
+        // density kg/m3
+        return Mw / vw;
+
+    }
+
+};
+
+template <class Scalar>
+const Scalar SimpleHuDuanH2O<Scalar>::R = Opm::Constants<Scalar>::R / 18e-3;
```
akva2, line 395: can't this just be static constexpr and given the value inline ?

opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
```diff
@@ -0,0 +1,427 @@
...
+    /*!
+     * \brief Returns the dynamic viscosity [Pa s] of the fluid phase given a set of parameters.
+     */
+    template <class Evaluation>
+    Evaluation viscosity(unsigned regionIdx,
+                         const Evaluation& temperature,
+                         const Evaluation& pressure,
+                         const Evaluation& /*Rs*/) const
+    {
+        //TODO: The viscosity does not yet depend on the composision
```
akva2, line 180: composition

akva2 (commented):
just some small drive-by things..

opm/material/components/SimpleHuDuanH2O.hpp
totto82: I guess so. But this is the way it's done for the other water components. I prefer not to change it in this file only.

totto82:
jenkins build this please

totto82:
I will merge this later today if nobody complains. It should not have any impact of the standard black-oil simulator.

PR #432, opened 2020-11-24
Do not check for kr curves for water single-phase

goncalvesmachadoc:
This fix remove the need of dummy relative permeability curves, so we can run flow_onephase and flow_onephase_energy with exactly input file as commercial simulator.


goncalvesmachadoc:
jenkins build this please

akva2:
jenkins build this please

atgeirr (commented):
The current changes are fine, but what happens in a single-phase oil-only case? A quick reading indicates to me that it should fail, but I might be wrong. We could check for single-phase condition at the top (just after the hasWater etc. variables have been created).

goncalvesmachadoc:
jenkins build this please

atgeirr:
Merging!

PR #433, opened 2020-12-07
Do not perform static initalization in headers.

atgeirr:
It seems to work with gcc but not with clang. A downstream PR will do the static init in opm-simulators.


akva2:
jenkins build this opm-simulators=2968 please

blattms:
I am wondering what "does not work" mean.

atgeirr:
> I am wondering what "does not work" mean.

The initializer for the static data will never be run. So running any CO2STOR case will fail upon first pvt evaluation.

blattms:
Thanks.

This whole inline usage looks a bit strange. I guess that is there to not create a library. 
Even if we make it work we might end up with multiple tables if we use the header in different compile units.

If we can do the instantiation at compile time, then  using [constexpr](https://pabloariasal.github.io/2020/01/02/static-variable-initialization/) might be a way to do this. Will cost compile time of course.

atgeirr:
Closing in favour of #435

PR #434, opened 2020-12-09
Extract Endpoints Using Common Facilities

bska:
This means we get the "SWCO" handling in SOGCR based on SGOF for free, and also removes duplicated and complex logic.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
@@ -176,28 +176,20 @@ public:
             enableKrnScaling_ = false;
             return;
         }
+
         // endpoint scaling is used, i.e., at least saturation scaling needs to be enabled
         enableSatScaling_ = true;
         enableThreePointKrSatScaling_ = endscale.threepoint();
 
-        if (twoPhaseSystemType == EclOilWaterSystem) {
-            // check if Leverett capillary pressure scaling is requested
-            if (eclState.getTableManager().useJFunc()) {
-                const auto& jfunc = eclState.getTableManager().getJFunc();
-                auto flag = jfunc.flag();
-                if (flag == Opm::JFunc::Flag::BOTH || flag == Opm::JFunc::Flag::WATER)
-                    enableLeverettScaling_ = true;
-            }
-        } else {
-            // check if Leverett capillary pressure scaling is requested
-            if (eclState.getTableManager().useJFunc()) {
-                const auto& jfunc = eclState.getTableManager().getJFunc();
-                auto flag = jfunc.flag();
-                if (flag == Opm::JFunc::Flag::BOTH || flag == Opm::JFunc::Flag::GAS)
-                    enableLeverettScaling_ = true;
-            }
-        }
+        if (eclState.getTableManager().useJFunc()) {
+            const auto flag = eclState.getTableManager().getJFunc().flag();
 
+            enableLeverettScaling_ = (flag == Opm::JFunc::Flag::BOTH)
+                || ((twoPhaseSystemType == EclOilWaterSystem) &&
+                    (flag == Opm::JFunc::Flag::WATER))
+                || ((twoPhaseSystemType == EclGasOilSystem) &&
+                    (flag == Opm::JFunc::Flag::GAS));
+        }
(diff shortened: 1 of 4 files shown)


bska:
This PR is has a downstream counterpart in OPM/opm-simulators#2974.  The two must be merged toghether.

totto82:
The code look good. Does it give the same endpoints?

opm/material/fluidmatrixinteractions/EclEpsConfig.hpp
```diff
@@ -212,11 +204,13 @@ class EclEpsConfig
...
         }
 
-        if (enablePcScaling_ && enableLeverettScaling_)
-            throw std::runtime_error("Capillary pressure scaling and the Leverett scaling function are "
-                                     "mutually exclusive: The deck contains the PCW/PCG property and the "
-                                     "JFUNC keyword applies to the water phase.");
-
+        if (enablePcScaling_ && enableLeverettScaling_) {
+            throw std::runtime_error {
+                "Capillary pressure scaling and the Leverett scaling function "
```
joakim-hove, line 209: I don't know the details here - but *maybe* the semi new functionality "prohibits" in the json config can be used to capture this at a very early stage. See [RTEMPVD](https://github.com/OPM/opm-common/blob/master/src/opm/parser/eclipse/share/keywords/000_Eclipse100/R/RTEMPVD) for an example.

bska:
> Does it give the same endpoints?

For the most part.  The one exception is "SOGCR" (critical oil saturation in oil/gas two-phase system) in the case of Family I keywords (SGOF/SWOF).  This PR accounts for the residual (connate) water saturation in deriving SOGCR, while the current master sources do not.  In other words, the current master sources interprets SGOF implicitly using the relation
```
So + Sg = 1
````
while this PR uses the relation
```
So + Sg + Swco = 1
```
to interpret the same two-phase saturation function data.  This PR **should** be more correct than the master sources, but there are several test cases for which this gives numerical results that don't match ECLIPSE as well as the current master sources.

I'm keeping this in draft state until I'm able to track down the differences.

**Comment by totto82** [comment]
> I'm keeping this in draft state until I'm able to track down the differences.

Cool. We did some testing many years on this but it was never tested thoroughly.

**Comment by bska** [comment]
jenkins build this opm-simulators=2974 please

**Comment by bska** [comment]
jenkins build this opm-simulators=2974 please

**Comment by bska** [comment]
jenkins build this opm-common=2177 opm-simulators=2974 please

**Comment by bska** [comment]
> I'm keeping this in draft state until I'm able to track down the differences.

I've added a hack that restores the SOGCR endpoint of the current master sources.  I hope to get to the bottom of what's happening, but in the meantime I'm going to mark this as "ready for review".

**Comment by bska** [comment]
jenkins build this opm-simulators=2974 please

**Comment by bska** [comment]
Unless anyone objects I intend to merge this PR along with its downstream OPM/opm-simulators#2974 no later than 10am CET tomorrow (Tuesday 5th January).

File: opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp [hunk 2]
```diff
@@ -156,105 +136,38 @@ struct EclEpsScalingPointsInfo
...
+        this->Sgl = rtep.connate.gas  [satRegionIdx];
+
+        this->Swcr  = rtep.critical.water       [satRegionIdx];
+        this->Sgcr  = rtep.critical.gas         [satRegionIdx];
+        this->Sowcr = rtep.critical.oil_in_water[satRegionIdx];
+        this->Sogcr = rtep.critical.oil_in_gas  [satRegionIdx];
+        if (family == SatFuncControls::KeywordFamily::Family_I) {
+            // Hack.  Papers over unknown problems elsewhere.
+            this->Sogcr += this->Swl;
         }
```
totto82, line 154: Could you add an "address" to elsewhere in case this HACK is long lived. It would make it possible to others to investigate this later.

bska: The trouble is I don't know where the issue is.  It is likely to be a combination of problems in multiple files.

bska:
Merging this now as advertised.

PR #435, opened 2020-12-10
Change static intialization of CO2 tables to avoid function call, use construction.

atgeirr:
This has been verified to work (unlike the current master) with clang on my system. I believe this is good to be merged when green.

Thanks to Arne Morten for the patch!

### DIFF
--- a/opm/material/common/UniformTabulated2DFunction.hpp
+++ b/opm/material/common/UniformTabulated2DFunction.hpp
@@ -62,6 +62,17 @@ public:
         resize(minX, maxX, m, minY, maxY, n);
     }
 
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

blattms:
Would be nice to know what the acually fix was. Is the move from `static` to `static const`/static constexpr` for the members of CO2Tables and the static initialization of them via inline or the move from arrays to vectors,

akva2:
As atgeirr wrote, the fix is moving from a function for initialization, to initialization through construction. The move to vector is mostly to work around annoying const double array not castable to const double**. As a bonus it makes the added constructor more versatile.

totto82 (approved):
Thanks for the fix!

PR #438, opened 2021-01-12
Fix Family I SOGCR Endpoint Hack

bska:
The maximum attainable oil saturation in the G/O system is `1-SWL`. Update call sites accordingly.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
+++ b/opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
@@ -186,8 +186,12 @@ public:
                                        Scalar& krnSwMdc,
                                        const Params& params)
     {
-        pcSwMdc = params.gasOilParams().pcSwMdc();
-        krnSwMdc = params.gasOilParams().krnSwMdc();
+        const auto Swco = params.Swl();
+
+        // Pretend oil saturation is 'Swco' larger than it really is in
+        // order to infer correct maximum Sg values in output layer.
+        pcSwMdc = std::min(params.gasOilParams().pcSwMdc() + Swco, Scalar{2.0});
+        krnSwMdc = std::min(params.gasOilParams().krnSwMdc() + Swco, Scalar{2.0});
 
         Valgrind::CheckDefined(pcSwMdc);
         Valgrind::CheckDefined(krnSwMdc);
@@ -203,8 +207,10 @@ public:
                                           const Scalar& krnSwMdc,
                                           Params& params)
     {
+        // Maximum attainable oil saturation is 1-SWL
+        const auto Swco = params.Swl();
         const double krwSw = 2.0; //Should not be used
-        params.gasOilParams().update(pcSwMdc, krwSw, krnSwMdc);
+        params.gasOilParams().update(pcSwMdc - Swco, krwSw, krnSwMdc - Swco);
     }
 
     /*!
@@ -220,7 +226,8 @@ public:
     static Evaluation pcgn(const Params& params,
                            const FluidState& fs)
     {
-        const auto Sw = 1.0 - Opm::decay<Evaluation>(fs.saturation(gasPhaseIdx));
+        // Maximum attainable oil saturation is 1-SWL.
+        const auto Sw = 1.0 - params.Swl() - Opm::decay<Evaluation>(fs.saturation(gasPhaseIdx));
         return GasOilMaterialLaw::twoPhaseSatPcnw(params.gasOilParams(), Sw);
     }
 
@@ -316,7 +323,8 @@ public:
     static Evaluation krg(const Params& params,
                           const FluidState& fluidState)
     {
-        const Evaluation Sw = 1.0 - Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
+        // Maximum attainable oil saturation is 1-SWL.
+        const Evaluation Sw = 1.0 - params.Swl() - Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
         return GasOilMaterialLaw::twoPhaseSatKrn(params.gasOilParams(), Sw);
     }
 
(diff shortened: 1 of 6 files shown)


bska:
Note: This PR depends on (and contains) #436 and must not be merged before that PR is approved and merged into the master sources.

bska:
I'm marking this as ready for review now.  The simulation results do differ from those in the regression suite, but local testing suggests the differences are not too big.  The NORNE_ATW2013 model reproduces in local test.  I will further assess the regression test failures on the Jenkins system.

bska:
jenkins build this opm-simulators=3014 please

opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
```diff
@@ -186,8 +186,12 @@ class EclDefaultMaterial : public TraitsT
...
                                        const Params& params)
     {
-        pcSwMdc = params.gasOilParams().pcSwMdc();
-        krnSwMdc = params.gasOilParams().krnSwMdc();
+        const auto Swco = params.Swl();
+
+        // Pretend oil saturation is 'Swco' larger than it really is in
+        // order to infer correct maximum Sg values in output layer.
+        pcSwMdc = std::min(params.gasOilParams().pcSwMdc() + Swco, Scalar{2.0});
+        krnSwMdc = std::min(params.gasOilParams().krnSwMdc() + Swco, Scalar{2.0});
```
totto82, line 195: I did not understand this. Can you explain more?

totto82 (approved):

opm/material/fluidmatrixinteractions/EclDefaultMaterial.hpp
totto82: The code changes looks ok. Some ref/const cleanups etc. 
I only looked at the code diff and have does not checked for missing '- Swl' 
This should be tested for the all the models. ("default* and stone1 and stone2) to make sure that no shifts are missed. 
Thanks for nesting this up and removing the Endpoint Hack. 
I have looked at some test failures and they seem small. Maybe we should ask @akva2 to make a "jenkins build this relaxed please" that runs the test with relaxed tolerances to focus on the real issues. That could potentially save us some time.

bska: > I did not understand this. Can you explain more?

I'll try.  For reasons of ECLIPSE compatibility we cannot output solution vectors named "KRNSW_OW", "KRNSW_GO", "PCSWM_OW" or "PCSWM_GO" ([link](https://github.com/OPM/opm-simulators/blob/3a0dbdc6e7adbd857d0371a13f291b065289c11e/ebos/ecloutputblackoilmodule.hh#L1056-L1066)) directly to the restart files.  Therefore, the output layer has a [hack](https://github.com/OPM/opm-common/blob/462ab5f44aa73baf57a408a2689f9204f603c62f/src/opm/output/eclipse/RestartIO.cpp#L545-L549) to translate these arrays to their (mostly) ECLIPSE compatible names, SOMAX and SGMAX.  That translation in turn assumes that
```
Sw = 1 - Sn
```
Adding `Swl` here ensures that that relation holds.  The `min(x, 2)` part is, admittedly, another hack to handle the initial condition for which the master sources return `Sn = 2` (translates to `Sw = -1` in the output layer).  I didn't want to go through yet another hoop to deal with that.

The proper solution here is **arguably** to create the ECLIPSE compatible hysteresis values right here and to remove the translation from the output layer.  I suggest doing so as a follow-up PR since I first need to check if this function is used in contexts other than outputting values to the restart files.

> This should be tested for the all the models. ("default* and stone1 and stone2) to make sure that no shifts are missed.

I've started that work locally.  The following cases in `opm-tests` activate `STONE1`, but I've not been able to find one that activates `STONE` (a.k.a., `STONE2`).  I'd be happy to be proven wrong.

* [model1/BASE1_MSW_HFA.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/BASE1_MSW_HFA.DATA#L227)
* [model1/BASE2_MSW_HFA.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/BASE2_MSW_HFA.DATA#L235)
* [model1/BASE3_MSW_HFA.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/BASE3_MSW_HFA.DATA#L234)
* [model1/BASE_MODEL_1.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/BASE_MODEL_1.DATA#L243)
* [model1/FAULTS_MODEL_1.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/FAULTS_MODEL_1.DATA#L255)
* [model1/MSW_MODEL_1.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/model1/MSW_MODEL_1.DATA#L243)
* [wsegsicd/TEST_WSEGSICD.DATA](https://github.com/OPM/opm-tests/blob/db18ed58167ffaf0c793491b8ef5cd66c6388e0f/wsegsicd/TEST_WSEGSICD.DATA#L198)

> Thanks for nesting this up and removing the Endpoint Hack.

Yeah, I was happy it turned out to be this limited in scope&mdash;assuming I've actually found all of them of course.

bska:
jenkins build this opm-simulators=3014 please

bska:
jenkins build this opm-simulators=3014 please

bska:
For what it's worth, I've checked the regression test failures of this (and #439) and we still capture the trends although there are visible differences in some of the summary curves for a small number of cases.  I am tempted to update the reference solutions and merge this now.

totto82:
> For what it's worth, I've checked the regression test failures of this (and #439) and we still capture the trends although there are visible differences in some of the summary curves for a small number of cases. I am tempted to update the reference solutions and merge this now.

Thanks for the huge effort of checking all the test failures. I support merging this and #439  as is with updated reference solutions. Small changes in the results are inevitable when changing core parts of the code.

bska:
> I support merging this and #439 as is with updated reference solutions.

Thanks.  I'll update the reference solutions for this PR first and then to #439 separately.

bska:
jenkins build this opm-simulators=3014 update_data please

bska:
jenkins build this opm-simulators=3014 opm-tests=473 please

bska:
New reference solutions have been merged into master.  Merging this now to bring the system back to a consistent state.

PR #440, opened 2021-01-13
fix the index into salinity in the BrineCo2Pvt model

totto82:
Thanks to valgrind


totto82:
jenkins build this please

totto82:
This is a trivial bugfix. I will self-merge tomorrow if nobody objects.

atgeirr:
> This is a trivial bugfix. I will self-merge tomorrow if nobody objects.

No need...

bska:
> This is a trivial bugfix. I will self-merge tomorrow if nobody objects.

I'll merge it right away myself.

PR #445, opened 2021-02-02
Explicitly Prevent Negative 3-p Kro Values in Stone 2 Model

bska:
This makes the material balance equations converge more smoothly in corner cases and makes a real field case run at all.


bska:
jenkins build this please

bska:
jenkins build this please

bska:
jenkins build this please

totto82:
This looks like a trivial fix. Whats the reason for keeping this a draft?

bska:
> Whats the reason for keeping this a draft?

Insufficient testing.  This change is known to improve the case added as OPM/opm-tests#399, but we don't have great test coverage of models with the `STONE2` keyword so I don't know if I'm breaking other models.

bska:
I'm marking this as "ready for review" now.  I've discussed the matter with Equinor collaborators and we agree that we can bring the fix into master despite the fact that we've not run exhaustive tests until now.

bska:
jenkins build this please

PR #446, opened 2021-02-02
Expose Two-Phase Oil Relperm As Individual Operation 

bska:
This PR adds new operations
```
relpermOilInOilGasSystem(params, fluidState)
relpermOilInOilWaterSystem(params, fluidState)
```
to the `Ecl*Material` implementations.  The primary objective is to support calculating the `KROG` and `KROW` quantities separately for summary output purposes (keywords `BKROG`, `BKROW`).  Reimplement the `krn` operation in terms of these new functions.

At present, these operations are supported only for three-phase simulation runs.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclStone1Material.hpp
+++ b/opm/material/fluidmatrixinteractions/EclStone1Material.hpp
@@ -359,11 +359,11 @@ public:
         // oil relperm at connate water saturations (with Sg=0)
         const Scalar krocw = params.krocw();
 
-        const Evaluation& Sw = Opm::decay<Evaluation>(fluidState.saturation(waterPhaseIdx));
-        const Evaluation& Sg = Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
+        const Evaluation Sw = Opm::decay<Evaluation>(fluidState.saturation(waterPhaseIdx));
+        const Evaluation Sg = Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
 
-        const Evaluation kro_ow = OilWaterMaterialLaw::twoPhaseSatKrn(params.oilWaterParams(), Sw);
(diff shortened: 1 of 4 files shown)


bska:
Note: This PR depends on and must not be merged before PR #445.

bska:
jenkins build this please

bska:
I'm marking this as "ready for review".

bska:
jenkins build this please

opm/material/fluidmatrixinteractions/EclStone1Material.hpp
```diff
@@ -359,11 +359,11 @@ class EclStone1Material : public TraitsT
         // oil relperm at connate water saturations (with Sg=0)
         const Scalar krocw = params.krocw();
 
-        const Evaluation& Sw = Opm::decay<Evaluation>(fluidState.saturation(waterPhaseIdx));
-        const Evaluation& Sg = Opm::decay<Evaluation>(fluidState.saturation(gasPhaseIdx));
+        const Evaluation Sw = Opm::decay<Evaluation>(fluidState.saturation(waterPhaseIdx));
```
totto82, line 362: This looks like a bug fix. What did Sw refer to here?

totto82 (approved):
Thanks. The code looks correct and should be merged.

opm/material/fluidmatrixinteractions/EclStone1Material.hpp
bska: > This looks like a bug fix.

More like a tiny optimisation.  `Opm::decay<>` constructs a new object and I recently became aware that there's no way for the compiler to perform lifetime extension (i.e., `const Evaluation& x = new_object`) without effectively doing something like this
```C++
const Evaluation __hidden = new_object;
const Evaluation& Sw = __hidden;
```
The above just bypasses the `__hidden` object.

PR #448, opened 2021-02-22
Extrapolate oil viscosity temperature factor for undersaturated case

totto82:
The linear extrapolation for oil viscosity temperature factor for the undersatured case is missing.


bska:
jenkins build this please

bska:
Do you have anything else to add here?  If not, I suggest we merge this in its current form.

totto82:
> Do you have anything else to add here? If not, I suggest we merge this in its current form.

I dont have anything more to add. I read your comment as an approval and will merge. Thanks.

bska:
> I read your comment as an approval

That was the intended meaning, yes.

PR #449, opened 2021-03-26
Stop pretending that we support CMake 2.8

blattms:
We actually already require at least CMake 2.8.12 due to the embedded pybind11 (some tests of it are even at 3.0). Anyway as Ubuntu LTS has 3.10.2 I doubt that anything less is tested by us.


blattms:
jenkins build this please

blattms:
Build check successful in OPM/opm-common#2386

PR #450, opened 2021-04-07
Use value objects for endpoint scaling

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2412


bska:
Merging to restore build following OPM/pom-common#2412.

PR #451, opened 2021-04-08
Correcting phase index.

osae:
Replacing two instances of non-wetting phase index with wetting phase index.


bska (approved):
Good catch!  I guess we've not used this model in a few years.

bska:
jenkins build this please

bska:
Build check is green.  I'll merge into master.

PR #452, opened 2021-04-13
add libpython3-dev to debian packaging

akva2:
needed since we now build embedded python support
in opm-common

Sorry for being late with this, so much going on atm (quarantine for my teenager, toddler at home, ..)


akva2:
this needs to be backported before first rc.

Issue #453, opened 2021-04-22
Official Debian Packages


blattms:
This is for people interested in this effort. The current state is on [salsa.debian.org](https://salsa.debian.org/blattms/opm-material). This is currently being finalized (mainly coming up with good package descriptions), and will then be submitted to debian.

blattms:
For updates of the process see OPM/opm-simulators#3172

PR #454, opened 2021-04-22
debian/control: Proposal for improved package descriptions

blattms:
For the proposal of OPM as a debian package (see issue #453). Comments/further improvements welcome.


blattms:
Closing as the plan should be to use the effort on https://salsa.debian.org/science-team/opm-material in the long run

PR #459, opened 2021-05-05
fixed: do not use Opm:: prefix within Opm namespace part 2

akva2:
have to use anchoring to root namespace in some places due
to overlapping namespace and class type names.

while at it fix some annoying issues in the pvt multiplexer classes.


akva2:
jenkins build this please

akva2:
jenkins build this opm-models=656 please

akva2:
jenkins build this opm-models=656 opm-simulators=3217 please

bska:
I've reviewed this now.  Looks good for the most part.  My only real concern is the `*Approach` enumerations being directly in the `Opm` namespace.  On the other hand, if that turns out to be a problem we can wrap them in a nested namespace later.  I'll merge this into master now.

PR #466, opened 2021-07-14
fixed use of wrong saturations for gas relperm for gas-water system

plgbrts:
Note: this PR is related to the twophase gas-water feature. A gas-water system however is not yet enabled/allowed.


totto82:
jenkins build this please

PR #467, opened 2021-07-28
Remove class EclHysteresisConfig - use Opm::EclHysterConfig

joakim-hove:
Was about to make some small validation inspired refactoring to the EHYSTR configuration in opm-common; remove this shim class to avoid double work.


joakim-hove:
jenkins build this please

PR #470, opened 2021-09-16
Extrapolate for values not in domain (UniformTabulated2DFunction).

blattms:
In the current code the interpolation actually already works if the values are outside of the tabulated region.

With this change there is now a static variable throwUntabulated (default: false) that we can set in opm-simulators. If it is false we will interpolate for every value instead of throwing and aborting (was the case always before).

### DIFF
--- a/opm/material/common/UniformTabulated2DFunction.hpp
+++ b/opm/material/common/UniformTabulated2DFunction.hpp
@@ -198,7 +198,6 @@ public:
     template <class Evaluation>
     Evaluation eval(const Evaluation& x, const Evaluation& y) const
     {
-#ifndef NDEBUG
         if (!applies(x,y))
         {
             std::string msg = "Attempt to get tabulated value for ("
@@ -216,7 +215,6 @@ public:
                 OpmLog::warning("PVT Table evaluation:" + msg + ". Will use extrapolation");
             }
         };
-#endif
 
         Evaluation alpha = xToI(x);
         Evaluation beta = yToJ(y);
--- a/opm/material/common/UniformTabulated2DFunction.hpp
+++ b/opm/material/common/UniformTabulated2DFunction.hpp
@@ -31,6 +31,7 @@
 #include <opm/material/common/Exceptions.hpp>
 #include <opm/material/common/MathToolbox.hpp>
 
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
FYI: Merging this will change the behaviour from throwing in UniformTabulated2dFunction to a warning.

Juts noticed that the #ifndef NDEBUG should probably be removed also. Agreed?

blattms:
jenkins build this please

opm/material/common/UniformTabulated2DFunction.hpp
```diff
@@ -186,21 +192,29 @@ class UniformTabulated2DFunction
      * \brief Evaluate the function at a given (x,y) position.
      *
      * If this method is called for a value outside of the tabulated
-     * range, a \c Opm::NumericalIssue exception is thrown.
+     * range, a \c Opm::NumericalIssue exception is thrown if
+     * throwUntabulated is true.
      */
     template <class Evaluation>
     Evaluation eval(const Evaluation& x, const Evaluation& y) const
     {
```
totto82, line 200: What about instead adding` bool extrapolate=false` like for the UnifromXTabulated2DFunction. That would make the behavior more similar and make it up to the call side to determine whether extrapolation is a wanted feature or not.

blattms: That is a good idea. Please be aware that at least for UniformXTabulated2DFunction, the call site is in opm-material (e.g. opm/material/fluidsystems/blackoilpvt/LiveOilPvt.hpp) where the value is actually hardcoded. Using the same approach we could limit the option of extrapolation to only Co2GasPvt. Should we go down this route?

If yes the static member would move to Co2GasPvt, set in the same as before in simulators, and that value would be used to call the eval function in UniformTabulated2DFunction.

Which of the other call sites should it be used? There are
- opm/material/eos/PengRobinson.hpp
- opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp

totto82: BrineCo2Pvt.hpp.

blattms:
Closing in favor of #473

PR #471, opened 2021-09-17
make oilReferenceDensity const

totto82:


totto82:
jenkins build this please

PR #472, opened 2021-09-28
warn not throw when brine and co2 properties are outside scope

totto82:


totto82:
Closing this in favor of #473

PR #473, opened 2021-09-28
Use extrapolation for unreasonable/untabulated values for CO2 sequestration cases.

blattms:
Better alternative to #470 which recommendation from @totto82 incorporated. With this the case now seems to run. No downstream change needed.

### DIFF
--- a/opm/material/common/UniformTabulated2DFunction.hpp
+++ b/opm/material/common/UniformTabulated2DFunction.hpp
@@ -202,7 +202,7 @@ public:
         {
             std::string msg = "Attempt to get tabulated value for ("
                 +std::to_string(double(scalarValue(x)))+", "+std::to_string(double(scalarValue(y)))
-                +") on a table of extend "
+                +") on a table of extent "
                 +std::to_string(xMin())+" to "+std::to_string(xMax())+" times "
                 +std::to_string(yMin())+" to "+std::to_string(yMax());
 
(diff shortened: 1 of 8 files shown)


blattms:
Not sure whether we really want to limit the number of warnings as in bc6a3da. Maybe more is better?

blattms:
jenkins build this please

totto82:
Thanks will test it. See also #472 which should take care of a similar issue for parametric functions.

totto82:
> Not sure whether we really want to limit the number of warnings as in bc6a3da. Maybe more is better?

I suggest to keep it the way it is first without limiting the number of warnings. If its getting annoying we can add something like you suggests.

blattms:
Just saw from your PR that opm-common is optional. Need to cater for that.

opm/material/common/UniformTabulated2DFunction.hpp
```diff
@@ -185,22 +186,31 @@ class UniformTabulated2DFunction
...
         if (!applies(x,y))
         {
-            throw NumericalIssue("Attempt to get tabulated value for ("
-                                   +std::to_string(double(scalarValue(x)))+", "+std::to_string(double(scalarValue(y)))
-                                   +") on a table of extend "
-                                   +std::to_string(xMin())+" to "+std::to_string(xMax())+" times "
-                                   +std::to_string(yMin())+" to "+std::to_string(yMax()));
+            std::string msg = "Attempt to get tabulated value for ("
+                +std::to_string(double(scalarValue(x)))+", "+std::to_string(double(scalarValue(y)))
+                +") on a table of extend "
```
bska, line 201: Possible follow-up here: Is this supposed to say "extent" instead?

blattms: not from me but fixes anyway

blattms:
jenkins build this please

totto82:
Thanks. Looks good and works.

PR #477, opened 2021-10-19
extrapolate binary coefficients for co2brine

totto82:
Adding some more extrapolation flags to make the CO2STORE option more robust.


blattms:
Lost my github-foo. Cannot find a way to review. Seem good though. just running a build check.

PR #478, opened 2021-10-20
extrapolate in calulateMoleFraction for CO2STORE

totto82:
Start using #477

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
@@ -435,7 +435,8 @@ private:
                                                     salinity_[regionIdx],
                                                     /*knownPhaseIdx=*/-1,
                                                     xlCO2,
-                                                    xgH2O);
+                                                    xgH2O,
+                                                    extrapolate);
 
         // normalize the phase compositions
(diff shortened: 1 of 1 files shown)


opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
```diff
@@ -435,7 +435,8 @@ class BrineCo2Pvt
                                                     salinity_[regionIdx],
                                                     /*knownPhaseIdx=*/-1,
                                                     xlCO2,
-                                                    xgH2O);
+                                                    xgH2O,
+                                                    extrapolate);
```
blattms, line 439: Mmh, isn't this new parameter unused and will just cause a warning? Did you miss something or does this fix compilation issues? Please clarify.

Anyway, please remove the new warning either by using the parameter or `[[maybe_unused]]`

totto82: I don't understand. Extrapolate which is true in BrineCo2Pvt will be passed to BinaryCoeffBrineCO2::calculateMoleFractions and overwrite the default i.e. extrapolate=false in that method.

blattms: :flushed: Should learn how to read code. Sorry, I misread it.

PR #481, opened 2021-11-02
Explicitly check for a == b before creating scale_factor = a/b

joakim-hove:
We have a case where the run halts with the following [assert](https://github.com/OPM/opm-simulators/blob/master/ebos/equil/equilibrationhelpers.hh#L834):
```C++
const PcEq<FluidSystem, MaterialLaw, MaterialLawManager> f(materialLawManager, phase, cell, targetPc);
double f0 = f(s0);
double f1 = f(s1);
if (f0 <= 0.0)
    return s0;
else if (f1 >= 0.0)
    return s1;
assert(f0 > 0 && f1 < 0);
```
the seemingly impossible assert triggers because `f0` and `f1` are both `nan`. The `nan` comes about due to the scaling function `EclEpsTwoPhaseLaw::unscaledToScaledPcnw_()` where the offending code is essentially:
```C++
double scaled_value = 0;
double unscaled_value = 0;
double alpha = scaled_value / unscaled_value;
```
The "fix" in this PR just recognizes the situation where `scaled_value == unsaled_value` and uses the value `alpha = 1` without actually evaluating the fraction. That is obviously a symptom fix, but I do not know enough about the domain here to evaluate whether there is an underlying bug to address here. For all it is worth this PR is enough to get the model to run through.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
@@ -482,7 +482,15 @@ private:
             return unscaledPcnw*alpha;
         }
         else if (params.config().enablePcScaling()) {
-            Scalar alpha = params.scaledPoints().maxPcnw()/params.unscaledPoints().maxPcnw();
+            const auto& scaled_maxPcnw = params.scaledPoints().maxPcnw();
+            const auto& unscaled_maxPcnw = params.unscaledPoints().maxPcnw();
+
+            Scalar alpha;
+            if (scaled_maxPcnw == unscaled_maxPcnw)
+                alpha = 1.0;
+            else
+                alpha = params.scaledPoints().maxPcnw()/params.unscaledPoints().maxPcnw();
+
             return unscaledPcnw*alpha;
         }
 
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
OK - with some assistance from Stein Krogstad I have looked more into this, and although still quite week on the subject I consider this a "quite good" fix and think it should be merged. The cell in question have specific PCG value of zero - and the manual can at least be interpreted as: "No scaling" in this case.

Unless there are objections I will merge this tomorrow.

opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp
```diff
@@ -482,7 +482,15 @@ class EclEpsTwoPhaseLaw : public EffLawT::Traits
             return unscaledPcnw*alpha;
         }
         else if (params.config().enablePcScaling()) {
-            Scalar alpha = params.scaledPoints().maxPcnw()/params.unscaledPoints().maxPcnw();
+            const auto& scaled_maxPcnw = params.scaledPoints().maxPcnw();
+            const auto& unscaled_maxPcnw = params.unscaledPoints().maxPcnw();
+
+            Scalar alpha;
+            if (scaled_maxPcnw == unscaled_maxPcnw)
+                alpha = 1.0;
```
totto82, line 490: Can we have a situation where unscaled is zero and scaled is not zero or is that not acceptable input?

joakim-hove: I don't know enough about this; my attempt was to "fix" this as non-invasive as possible. The situation you describe could/should be handled with an additonal:

```C++
if (!std::isfinite(alpha))
     throw std::logic_error("Invalid numerical value encountered ....");
```

happy to add that. But as I have said numerous times - I am careful in this part of town.

PR #482, opened 2021-11-04
changed: avoid manual memory management

akva2:
See no reason not to use a std::vector here. My main motivation is it causes some (false) static analyzer issues.

### DIFF
--- a/opm/material/common/FastSmallVector.hpp
+++ b/opm/material/common/FastSmallVector.hpp
@@ -33,6 +33,7 @@
 
 #include <array>
 #include <algorithm>
+#include <vector>
 
 namespace Opm {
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

atgeirr:
> See no reason not to use a std::vector here.

I can see one: This increases the minimum memory usage of the class significantly. So if a large array of these is used it will be costly. But I think they are only used a few at a time in the well code, so it should be ok.

atgeirr:
jenkins build this please

bska:
> > See no reason not to use a `std::vector` here.
> 
> I can see one: This increases the minimum memory usage of the class significantly.

If we're concerned about size there's always the option of using `std::unique_ptr<T[]>` instead.  That class has `operator[]()` and knows how to call `delete[]` instead of `delete` to dispose of the memory.  We still need a custom destructor if the `T` is not [trivially destructible](https://en.cppreference.com/w/cpp/types/is_destructible) though

atgeirr:
benchmark please

ytelses:
Benchmark result overview: 




| Test    | Configuration                                     | Relative |
| ------- | ------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.011    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.018    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.01     |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.008    |
* Speed-up = Total time master / Total time pull request. Above 1.0 is an improvement. *



View result details @ https://www.ytelses.com/opm/?page=result&id=1729

PR #488, opened 2021-12-14
changed: avoid long vectors of shared ptrs

akva2:
this use a lot of memory with many elements.

For a case I'm looking at this brought memory usage down from
58.9GB (61GB peak) to 47.3GB (which is also the peak).

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLawParams.hpp
+++ b/opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLawParams.hpp
@@ -106,8 +106,8 @@ public:
     /*!
      * \brief Set the scaling points which are seen by the physical model
      */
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

alfbr:
benchmark please

blattms:
I guess the numbers are for a serial run or are we doing the computatioj on the unloadbalanced grid?

akva2:
it's all serial.

akva2:
Added some more commits that brings it down to 46.5949GB. Another big gain can be had by disabling the non-conditional diffusion storage in the intensive quantities, but that is separate.

akva2:
jenkins build this please

alfbr:
These are very impressive gains! Needless to say, we really look forward to seeing these improvements in master :-)

atgeirr:
Read through, looks good, merging!

PR #489, opened 2021-12-29
Saturation functions based on LET-curves.

osae:
Includes multiplexer for run-time selection of LET vs standard tabulated functions.
This PR depends on PR https://github.com/OPM/opm-common/pull/2917

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
+++ b/opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
@@ -31,8 +31,8 @@
 #ifndef OPM_ECL_MATERIAL_LAW_MANAGER_HPP
 #define OPM_ECL_MATERIAL_LAW_MANAGER_HPP
 
+#include <opm/material/fluidmatrixinteractions/SatCurveMultiplexer.hpp>
 #include <opm/material/fluidmatrixinteractions/EclTwoPhaseMaterialParams.hpp>
-#include <opm/material/fluidmatrixinteractions/PiecewiseLinearTwoPhaseMaterial.hpp>
 #include <opm/material/fluidmatrixinteractions/EclEpsTwoPhaseLaw.hpp>
 #include <opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLaw.hpp>
 #include <opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp>
@@ -80,9 +80,9 @@ private:
     typedef TwoPhaseMaterialTraits<Scalar, waterPhaseIdx, gasPhaseIdx> GasWaterTraits;
 
     // the two-phase material law which is defined on effective (unscaled) saturations
-    typedef PiecewiseLinearTwoPhaseMaterial<GasOilTraits> GasOilEffectiveTwoPhaseLaw;
-    typedef PiecewiseLinearTwoPhaseMaterial<OilWaterTraits> OilWaterEffectiveTwoPhaseLaw;
-    typedef PiecewiseLinearTwoPhaseMaterial<GasWaterTraits> GasWaterEffectiveTwoPhaseLaw;
+    typedef SatCurveMultiplexer<GasOilTraits> GasOilEffectiveTwoPhaseLaw;
+    typedef SatCurveMultiplexer<OilWaterTraits> OilWaterEffectiveTwoPhaseLaw;
+    typedef SatCurveMultiplexer<GasWaterTraits> GasWaterEffectiveTwoPhaseLaw;
 
     typedef typename GasOilEffectiveTwoPhaseLaw::Params GasOilEffectiveTwoPhaseParams;
     typedef typename OilWaterEffectiveTwoPhaseLaw::Params OilWaterEffectiveTwoPhaseParams;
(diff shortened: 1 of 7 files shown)


opm/material/fluidmatrixinteractions/SatCurveMultiplexer.hpp
```diff
@@ -0,0 +1,383 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
```
blattms, line 8: I guess this should be 3 for new code

bska:
I haven't looked at this at all, but is this feature supposed to interact with end-point scaling (vertical or horizontal) in any way?

bska:
I'd **really** prefer not to have multiplexing here.  Is there no other way?

totto82:
jenkins build this opm-common=2917

totto82:
jenkins build this opm-common=2917 please

atgeirr:
Although this PR is large, the changes that are most critical to evaluate are those that modify Flow behaviour. The intention here is to preserve existing behaviour when there are no LET input, but we must check that carefully. I therefore intend to review the changes to `EclMaterialLawManager` as well as the new multiplexer thoroughly, but assume that the new LET functionality is well enough tested to be mergeable (as an experimental feature at worst), as that code should not have any influence at all on existing cases. Similarly, the opm-common PR should be vetted in that it should not allow new failures related to missing or mismatching keywords etc. compared to the current master. Due to current workload I will probably not do these reviews until after Easter, so I appreciate dropping the release milestone.

opm/material/fluidmatrixinteractions/TwoPhaseLETCurves.hpp
```diff
@@ -0,0 +1,312 @@
...
+    //! are dependent on the phase composition
+    static const bool isCompositionDependent = false;
+
+    /*!
+     * \brief The capillary pressure-saturation curves.
+     */
+    template <class Container, class FluidState>
+    static void capillaryPressures(Container& /* values */, const Params& /* params */, const FluidState& /* fs */)
+    {
+        throw std::invalid_argument("The capillaryPressures(fs) method is not yet implemented");
```
atgeirr, line 98: All of these should throw std::logic_error instead of invalid_argument.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
```diff
@@ -741,6 +741,45 @@ class EclMaterialLawManager
...
+
+                // S=(So-Sogcr)/(1-Sogcr-Sgcr-Swco),  krog = Krt*S^L/[S^L+E*(1.0-S)^T]
+                const Scalar s_min_w = letSgofTab.s2_critical;
+                const Scalar s_max_w = 1.0-letSgofTab.s1_critical-Swco;
+                const std::vector<Scalar>& letCoeffsOil = {s_min_w, s_max_w,
+                                                           static_cast<Scalar>(letSgofTab.l2_relperm),
+                                                           static_cast<Scalar>(letSgofTab.e2_relperm),
+                                                           static_cast<Scalar>(letSgofTab.t2_relperm),
+                                                           static_cast<Scalar>(letSgofTab.krt2_relperm)};
+                realParams.setKrwSamples(letCoeffsOil, dum);
```
atgeirr, line 759: I think this approach works, but I cannot say I find it easy to understand. I am here talking about copying the data to a dumb vector, passing it through an interface that has no obvious connection to what is really going on, and unpacking on the inside.

Unless my brain is fooling me, at this point the type of `realParams` should be `ThoPhaseLETCurvesParam`, and then you could make one or a few interface methods for it that are reasonably named and has properly named parameters etc.

atgeirr (commented):
This can be merged. I do not think there is a simple way to do this with e.g. virtual functions instead since the functions are templates, and their parameters are of different classes. However, I think the way data are passed into the parameter objects can be simplified a lot, see the comment.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
osae: The idea was to adapt to the interface already used by the `PiecewiseLinearTwoPhaseMaterial`. To my understanding, introducing tailored methods for the `TwoPhaseLETCurvesParams` would have to impact `PiecewiseLinearTwoPhaseMaterialParams` as well due to the multiplexer construction.

opm/material/fluidmatrixinteractions/TwoPhaseLETCurves.hpp
osae: Fixed.

opm/material/fluidmatrixinteractions/EclMaterialLawManager.hpp
osae: Supplied a comment to `dum` hinting at what is going on.

totto82:
jenkins build this opm-common=2917 please

totto82:
> This can be merged.
The minor comments has been addressed. I will merge now.

PR #493, opened 2022-04-01
Converted BrineCo2Pvt.hpp to UTF-8

blattms:


opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
```diff
@@ -352,7 +352,7 @@ class BrineCo2Pvt
         Scalar M_CO2 = CO2::molarMass();
         Scalar M_H2O = H2O::molarMass();
 
-        const LhsEval& tempC = temperature - 273.15;        /* tempC : temperature in �C */
+        const LhsEval& tempC = temperature - 273.15;        /* tempC : temperature in °C */
```
bska, line 355: I doesn't show in the GitHub review interface, but I **assume** you've only changed the encoding of the degrees Centigrade symbol here, right?

bska, line 355: I doesn't show in the GitHub review interface, but I **assume** you've only changed the encoding of the degrees Centigrade symbol here, right?

totto82:
Thanks have notices the issue in the master and trust you that this is a fix. Will merge....

blattms:
Thanks. iconv should only have changed what was needed (read single characters). Hence I assume you are right.

PR #497, opened 2022-04-11
Add PVT for combined wet (vaporized oil)  and humid (vaporized water) gas

plgbrts:
Both PVT tables `PVTG `and `PVTGW` are used to specify the pvt for wet&humid gas with the assumption that the properties (viscosity `mu` and formation volume factor `Bg`) are evaluated for table  
- PVTG at saturated water-gas ratios (`Rvw_sat`)
- PVTGW at saturated oil-gas ratios (`Rv_sat`)

For example in case gas-oil co-exist and water has disappeared  the propertie values can be retrieved from the PVTGW table that contains data for saturated Rv (i.e. `Rv_sat`) and undersaturated Rvw

For the special case that only gas is present and hence `Rvw` and `Rv` are undersaturated we use the approximation `Bg(Rv,Rvw,p) = Bg(Rv,Rvw_sat,p)`, (and similar for `mu`) for which the values are then retrieved from the `PVTG` table.

### DIFF
--- a/examples/co2brinepvt.cpp
+++ b/examples/co2brinepvt.cpp
@@ -38,7 +38,8 @@ double densityGas(const Co2Pvt& co2Pvt, const double p, const double T, const do
     return co2Pvt.inverseFormationVolumeFactor(/*regionIdx=*/0,
                                                   T,
                                                   p,
-                                                  Rv) * co2Pvt.gasReferenceDensity(0);
+                                                  Rv,
+                                                  /*Rvw=*/0.0) * co2Pvt.gasReferenceDensity(0);
 }
 
 template <class BrinePvt>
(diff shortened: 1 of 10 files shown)


totto82:
Thanks for the contribution. I think this is a valuable extension of Flow . I have looked at the change and have two initial comments. 
1) Currently the pvt model interface does not have a method that return the density directly. (only the Fluidsystem has) When phase specific density is needed in the initialization and in the well model the b factor ++ is used to compute the density. Until now that has worked ok. But with the addition of rvw the density computation becomes more complicated. (and even more if a rsw is added) And its easy to forget to adjust for the rs/rv/rvw when computing the density. (I think with the current PR in opm-models and opm-simulators some times computes the wrong density because the rvw effect is not accounted for.) But rather than fix those places, my suggestion would be to add a density method to the pvt models. That should be done in another PR before this. I can do that work, but then this PR needs to be rebased on top of that. What do you think @plgbrts, and @bska 
2) The second point is the primary-variable switching part of the code in opm-models. Again, I think your solution is ok as you follow the current structure of the code. But with more switching options the code and logic becomes more complicated. I don't have a good solution for this at the moment, but maybe we should take the time to discuss this part of the code a bit more in detail in a video-meeting. Anybody wants to join?

atgeirr:
> Anybody wants to join?

Count me in.

plgbrts:
jenkins build this opm-models=688 opm-simulators=3869 please

totto82:
jenkins build this opm-models=688 opm-simulators=3869 update_data please

bska:
jenkins build this opm-models=688 opm-simulators=3869 opm-tests=764 please

PR #498, opened 2022-04-12
Fix conversion from density to bo

totto82:
Correctly account for dissolved CO2 in the conversion from density to bo

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/BrineCo2Pvt.hpp
@@ -209,7 +209,7 @@ public:
                                             const Evaluation& pressure,
                                             const Evaluation& Rs) const
     {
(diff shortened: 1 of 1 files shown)


totto82:
This fixes a critical bug and should be ported to the release branch.

totto82:
jenkins build this please

blattms (approved):
Looks good to me. (albeit I am lacking the knowledge to really tell)
I guess the test failures are expected because of the changes?

totto82:
> I guess the test failures are expected because of the changes?

Actually not, since it should not affect cases where rs = 0
I will do some more checking....

totto82:
> I will do some more checking....

There seems to be a bug in the initialization code in master https://github.com/OPM/opm-simulators/pull/3871

totto82:
jenkins build this please

totto82:
Now the test failures are expected. I will update the ref.

totto82:
jenkins build this update_data please

blattms:
jenkins build this opm-tests=749 please

PR #501, opened 2022-04-14
[Release 2022.04] Fix bug WATDENT 

goncalvesmachadoc:
The implementation of WATDENT is not retrieving the values for Pref and Bw from the PVTW table, making a division by zero in the inverseFormationVolumeFactor() computation:
return 1.0/(((1 - X)*(1 + cT1*Y + cT2*Y*Y))*BwRef);
and breaking the simulation run.

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/WaterPvtThermal.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/WaterPvtThermal.hpp
@@ -138,6 +138,14 @@ public:
                 watdentCT1_[regionIdx] = record.C1;
                 watdentCT2_[regionIdx] = record.C2;
             }
+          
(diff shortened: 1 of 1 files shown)


blattms:
Thanks, would be cool if yiu added some explanation (e.g, what bug) to the commit message. That might help others.

goncalvesmachadoc:
The implementation of WATDENT is not retrieving the values for Pref and Bw from the PVTW table, making a division by zero in the inverseFormationVolumeFactor() computation:
` return 1.0/(((1 - X)*(1 + cT1*Y + cT2*Y*Y))*BwRef);`
and breaking the simulation run.

blattms:
jenkins build this please

PR #504, opened 2022-04-20
return rsSat = 0 when dissolution is disabled in co2module

totto82:
See discussion OPM/opm-simulators#3871


totto82:
jenkins build this please

atgeirr:
Thanks, merging!

PR #505, opened 2022-04-21
only remove duplicated points if there are more than two unique points

totto82:
https://github.com/OPM/opm-simulators/issues/3881

### DIFF
--- a/opm/material/fluidsystems/blackoilpvt/DryHumidGasPvt.hpp
+++ b/opm/material/fluidsystems/blackoilpvt/DryHumidGasPvt.hpp
@@ -643,7 +643,8 @@ private:
         //Prune duplicate Rv values (can occur, and will cause problems in further interpolation)
         auto x_coord_comparator = [](const Pair& a, const Pair& b) { return a.first == b.first; };
         auto last = std::unique(pSatSamplePoints.begin(), pSatSamplePoints.end(), x_coord_comparator);
-        pSatSamplePoints.erase(last, pSatSamplePoints.end());
+        if (std::distance(pSatSamplePoints.begin(), last) > 1) // only remove them if there are more than two points
+            pSatSamplePoints.erase(last, pSatSamplePoints.end());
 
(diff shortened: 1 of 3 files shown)


totto82:
jenkins build this please

totto82:
@bska Any comments? Is this is a valid/sufficient fix for the issue? We may want to consider including this in the release. I leave that up to @bska to decide.

opm/material/fluidsystems/blackoilpvt/DryHumidGasPvt.hpp
```diff
@@ -643,7 +643,8 @@ class DryHumidGasPvt
         //Prune duplicate Rv values (can occur, and will cause problems in further interpolation)
         auto x_coord_comparator = [](const Pair& a, const Pair& b) { return a.first == b.first; };
         auto last = std::unique(pSatSamplePoints.begin(), pSatSamplePoints.end(), x_coord_comparator);
-        pSatSamplePoints.erase(last, pSatSamplePoints.end());
+        if (pSatSamplePoints.begin() + 1 != last) // only remove them if there are more than two points
```
bska, line 646: I think I'd prefer to phrase this condition in terms of [`std::distance()`](https://en.cppreference.com/w/cpp/iterator/distance) instead.  The expression is **probably** well-defined behaviour as written, but `std::distance()` has fewer pitfalls:
```C++
if (std::distance(x.begin(), last) > 1) {
    // Prune duplicate elements if more than two unique elements in sequence
}
```

totto82: Thanks.

totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

bska:
Since the build check is green now I think we'll take this into the master branch.  I'm not comfortable taking it into the release branch however.  This change needs more testing on "unexpected" cases before we declare that it's ready for production use.

PR #509, opened 2022-04-27
Fix installation of man page.

blattms:
The documentation directory is called doc here, not docs (likein another module).


blattms:
jenkins build this please

blattms:
A bit strange that CMake did not complain about the wrong directory. Maybe I just missed it,

PR #515, opened 2022-06-20
Extrapolate gas diffusion in CO2 module

totto82:
Add missing "extrapolate  = true" when calling `binaryCoeffBrineCO2::gasDiffCoeff`


totto82:
jenkins build this please

PR #516, opened 2022-06-27
PT Flash for multiple components

trinemykk:
Added the isobaric-isothermal equilibrium (PT Flash) to constraint solvers, work done together with Svenn, Kai and Halvor. This is compared to implementations in MRST and Julia by Olav Moyner. Two tests, one standard three components and one with co2 and brine as two components are added. The more classical cubic root finder is implemented to find the needed roots, and the derivatives is now handled through the implicit function theorem. Some of the components have criticalVolume() and acentricFactor(), that is needed in this flash. Have also added opm/material/viscositymodels which currently contain two viscositymodels: the standard LBC from (Lohrenz, Bray and Clark 1964), and the modified LBC more specific for CO2rich mixtures (Lansangan, Taylor, Smith & Kovarik 1993). Thanks to Ove for the viscosity implementation.

PTFlash will be improved, so this can be looked as a draft PR.


trinemykk:
@totto82  and @atgeirr please take a look

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

atgeirr:
The obvious question is: How should we go about reviewing and merging this? Going through all the code (~1200 loc for PTFlash, plus ~1800 loc in other classes) in detail is a monumental task. On the other hand, it seems to me that most of this code has already been seen by and/or worked on by multiple persons.

So: what aspects of the code should be looked at, tested or reviewed before we press the green button? What classes or functions can be considered “already reviewed”? Do the two tests cover all the code?

bska:
benchmark please

opm/material/fluidsystems/Spe5ParameterCache.hpp
```diff
@@ -175,6 +175,25 @@ class Spe5ParameterCache
...
+     */
+    Scalar aCache(unsigned phaseIdx, unsigned compIdx, unsigned compJIdx)
+    {
+        switch (phaseIdx)
+        {
+        case oilPhaseIdx: return oilPhaseParams_.getaCache(compIdx,compJIdx);
+        case gasPhaseIdx: return gasPhaseParams_.getaCache(compIdx,compJIdx);
+        default:
+            throw std::logic_error("The aCache() parameter is only defined for "
+                                   "oil phase");
```
bska, line 193: Should this diagnostic message mention the gas phase too?  There does at least seem to be code to handle the gas phase here.

opm/material/viscositymodels/LBC.hpp
```diff
@@ -0,0 +1,213 @@
...
+            }
+            my0 += xrM*mys;
+            sumxrM += xrM;
+            }
+            my0 /= sumxrM;
+
+            std::vector<Scalar> LBC = {0.10230,
+                                   0.023364,
+                                   0.058533,
+                                   -0.040758,  // trykkfeil i 1964-artikkel: -0.40758
```
bska, line 199: Translate comment into English?

opm/material/viscositymodels/LBCmodified.hpp
```diff
@@ -0,0 +1,133 @@
...
+            }
+            my0 += xrM*mys;
+            sumxrM += xrM;
+            }
+            my0 /= sumxrM;
+
+            std::vector<Scalar> LBC = {0.10230,
+                                   0.023364,
+                                   0.058533,
+                                   -0.040758,  // trykkfeil i 1964-artikkel: -0.40758
```
bska, line 119: Comment in English, please.

tests/test_co2brine_ptflash.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // TODO: here the zInit does not have the proper derivatives
+    Flash::solve(fs, zInit, spatialIdx, flash_verbosity, flash_twophase_method, flash_tolerance);
+
+}
+
+int main(int argc, char **argv)
+{
+    Dune::MPIHelper::instance(argc, argv);
+
+    testCo2BrineFlash();
```
bska, line 143: This test appears to be restricted to "compile + run" only.  How hard would it be to add checks for "produces expected result" too?

tests/test_threecomponents_ptflash.cpp
```diff
@@ -0,0 +1,139 @@
...
+    using Flash = Opm::PTFlash<double, FluidSystem>;
+    Flash::solve(fluid_state, z, spatialIdx, flash_verbosity, flash_twophase_method, flash_tolerance);
+
+}
+
+int main(int argc, char **argv)
+{
+    Dune::MPIHelper::instance(argc, argv);
+
+    testPTFlash();
```
bska, line 136: Is this test restricted to "compile + run" only?

opm/material/common/PolynomialUtils.hpp
```diff
@@ -309,6 +309,105 @@ unsigned invertCubicPolynomial(SolContainer* sol,
 
     return 3;
 }
+
+/*!
+ * \ingroup Math
+ * \brief Invert a cubic polynomial analytically
```
bska, line 315: Just out of curiosity, why do we need an exact cubic solver instead of something resembling a Newton loop or eigenvalues of the companion matrix for that matter?

bska (commented):
I've only looked at the most superficial part of this PR&mdash;diagnostic messages &c&mdash;so this is not a thorough review.  Still, there are a couple of issues that I'd like to see being addressed before I'm prepared to merge this.

ytelses:
Benchmark result overview: 




| Test    | Configuration                                     | Relative |
| ------- | ------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.004    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.004    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 1.006    |
* Speed-up = Total time master / Total time pull request. Above 1.0 is an improvement. *



View result details @ https://www.ytelses.com/opm/?page=result&id=1693

trinemykk:
> The obvious question is: How should we go about reviewing and merging this? Going through all the code (~1200 loc for PTFlash, plus ~1800 loc in other classes) in detail is a monumental task. On the other hand, it seems to me that most of this code has already been seen by and/or worked on by multiple persons.
> 
> So: what aspects of the code should be looked at, tested or reviewed before we press the green button? What classes or functions can be considered “already reviewed”? Do the two tests cover all the code?

@atgeirr , my view on this is the following:

0) the main contribution is the file _opm/material/constraintsolvers/PTFlash.hpp_,
this is the actual flash where the whole procedure is done (including stability test, rachford rice, ssi, newton, handling of the derivatives..) - the results from the substeps (and the end result) is manually checked against the test in [git@github.com:moyner/MultiComponentFlash.jl.git] ,  for two fluidsystems _opm/material/fluidsystems/ThreeComponentFluidSystem.hh_ and _opm/material/fluidsystems/Co2BrineFluidSystem.hh_ through the tests  _tests/test_co2brine_ptflash.cpp_ and _tests/test_threecomponents_ptflash.cpp_ , 
so the functionality of this code can in my opinion be merged directly (but the technical coding can definitely be improved, but I am not sure this is important at this point)
A necessary complementary file is opm/material/fluidsystems/PTFlashParameterCache.hpp, that is very similar to the other ParameterCache-files found in the directory

1) the files that are modified with minor things (and can be merged directly) are:
CMakeLists.txt, opm/material/common/MathToolbox.hpp, opm/material/components/Brine.hpp, opm/material/components/Component.hpp, opm/material/components/H2O.hpp, opm/material/components/N2.hpp, 
opm/material/components/N2.hpp, opm/material/components/SimpleCO2.hpp, opm/material/components/TabulatedComponent.hpp, opm/material/components/iapws/Common.hpp, opm/material/densead/Math.hpp, opm/material/fluidsystems/BaseFluidSystem.hpp, opm/material/fluidsystems/Spe5ParameterCache.hpp, tests/checkComponent.hpp, opm/material/fluidstates/FluidStateCompositionModules.hpp

2) the files that are modified for the purpose of getting the flash to work are:
_opm/material/eos/PengRobinson.hpp_,  _opm/material/eos/PengRobinsonMixture.hpp_, _opm/material/eos/PengRobinsonParamsMixture.hpp_, 
since the former version of these files didn't produce the wanted roots of the cubic polynomial solved through the stability test and different derivatives on one of the equations (compared to the test from Moyner in 0)), it is safe to merge this directly. The former version of these files has never been tested and verified against other simulators like we have done now.

3) the files we have added as extra functionality needed are:
the viscositymodels: _opm/material/viscositymodels/LBC.hp_p, _opm/material/viscositymodels/LBCmodified.hpp_, these are "standard" methods found in the literature and the results produced are manually compared against viscosities produced by 
[git@github.com:moyner/MultiComponentFlash.jl.git], and this can in my opinion be merged when the Norwegian comments are fixed as @bska pointed out

4) the specific fluidsystems tested here needs the added components _opm/material/components/C1.hpp_, _opm/material/components/C10.hpp_, (nothing controversial here), and another component added since we had it here was 
_opm/material/components/H2.hpp_ (not needed by the tests, but nice to have in place), can be merged directly

5) In the _file opm/material/common/PolynomialUtils.hpp_ a more standard cubic root solver is added _cubicRoots()_, this one produces the same roots as the compared code - so this functionality should also be safe to merge 

Summary:
-  the flash produced by the two fluidsystems provided is manually tested against [git@github.com:moyner/MultiComponentFlash.jl.git], and should simplify the review process
-  especially the file PTFlash has improvements to be made when it comes to efficiency, compactness, and maybe language -  but this does not affect other parts of the code and if there his nothing extremely ugly it can slide (?)
-  all files except the LBCmodified.hpp and the components H2.hpp is involved in the test,
but some methods (especially in PTFlash is not used in these tests, since both are in the two phase regime), but I will shortly push a test that involves a single phase case where the rest of the methods are used (have tested this on a problem in opm-models)

opm/material/common/PolynomialUtils.hpp
moyner: Newton requires initial guesses and tolerances while the explicit roots are straightforward to compute and crucially allows AD to pass straight through which is more tricky if we call a algorithm.

tests/test_co2brine_ptflash.cpp
GitPaean: Yeah, we should be able to check the results.

opm/material/viscositymodels/LBC.hpp
```diff
@@ -0,0 +1,213 @@
...
+            sumLBC += Opm::pow(rho_r,i)*LBC[i];
+        }
+
+        return (my0 + (Opm::pow(sumLBC,4.0) - 1e-4)/zeta_tot)/1e3; // mPas-> Pas
+    }
+
+
+        // Improved LBC model for CO2 rich mixtures. (Lansangan, Taylor, Smith & Kovarik - 1993)
+        template <class FluidState, class Params, class LhsEval = typename FluidState::Scalar>
+        static LhsEval LBCmodified(const FluidState& fluidState,
```
GitPaean, line 126: If I did not see wrong, there are two `LBCModified` functions now, and one should be removed. 

And also, `LBCmodified` needs a more specific name.

opm/material/fluidsystems/Co2BrineFluidSystem.hh
```diff
@@ -0,0 +1,190 @@
...
+         * \copydoc BaseFluidSystem::viscosity
+         */
+        template <class FluidState, class LhsEval = typename FluidState::Scalar, class ParamCacheEval = LhsEval>
+        static LhsEval viscosity(const FluidState& fluidState,
+                                 const ParameterCache<ParamCacheEval>& paramCache,
+                                 unsigned phaseIdx)
+        {
+            // Use LBC method to calculate viscosity
+            LhsEval mu;
+            mu = ViscosityModel::LBC(fluidState, paramCache, phaseIdx);
```
GitPaean, line 166: Following the discussion before, I think here the `LBCModified` should be used, right?

trinemykk: Both can be used, when comparing to the Julia code the standard LBC is used

opm/material/viscositymodels/LBC.hpp
trinemykk: I will fix this

GitPaean:
Just a note. Besides the single phase situation, we should also test all the three methods, 
```
ssi , newton, ssi+newton
```
I can help some.

opm/material/fluidsystems/Spe5ParameterCache.hpp
trinemykk: yes, will add the gas phase

opm/material/common/PolynomialUtils.hpp
bska: > crucially allows AD to pass straight through which is more tricky if we call an algorithm.

Right, I didn't think of that aspect.  I suppose another alternative would be [Laguerre's method](https://en.wikipedia.org/wiki/Laguerre%27s_method) with deflation, but since we know we're solving a cubic polynomial using the exact formulae is a reasonable strategy despite the messy calculations.

atgeirr:
Thanks @trinemykk for your extensive explanation. After reading this I agree with you: The only modification of existing behaviour is for things (PengRobinson etc.) that were not working well, so it is fine to merge it from a behaviour point of view.

From a testing p.o.v. I think the additions discussed above (single-phase, and all three flash solution methods) sounds sufficient.

Finally, code quality: I agree that it is better to have something that works than to have nothing... While I have not looked at the code very much myself, knowing that it has been written in collaboration by at least two persons should ensure a sufficient level of quality.

opm/material/fluidsystems/Co2BrineFluidSystem.hh
GitPaean: Currently, LBCCo2Rich is not used anywhere. Maybe you can set a flag to `Co2BrineFluidSystem` to control whether `LBC` or `LBCCo2Rich` is used for the viscosity calcualtion?

atgeirr:
jenkins build this please

tests/test_threecomponents_ptflash.cpp
GitPaean: I added some checking of the values and derivatives after the flash calculation, so the results of the flash calculation is checked.

GitPaean:
jenkins build this please

tests/test_co2brine_ptflash.cpp
```diff
@@ -0,0 +1,328 @@
...
+    for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
+        const Evaluation Ktmp = fluid_state.wilsonK_(compIdx);
+        fluid_state.setKvalue(compIdx, Ktmp);
+    }
+    const Evaluation Ltmp = 1.;
+    fluid_state.setLvalue(Ltmp);
+
+    const int spatialIdx = 0;
+    using Flash = Opm::PTFlash<double, FluidSystem>;
+    Flash::solve(fluid_state, z, spatialIdx, flash_verbosity, flash_twophase_method, flash_tolerance);
```
GitPaean, line 287: Sorry, I changed the interface of the function a little bit, so this line should be 
```c++
Flash::solve(fluid_state, z, spatialIdx, flash_twophase_method, flash_tolerance, flash_verbosity);
```

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

atgeirr:
Thank you for your effort, I will merge this now!

PR #521, opened 2022-08-02
Fixes in some component classes

akva2:
Modernization (use constexpr where appropriate) and avoid use of dangling references.

### DIFF
--- a/opm/material/components/Brine.hpp
+++ b/opm/material/components/Brine.hpp
@@ -161,10 +161,10 @@ public:
             { +0.17965e-2, +0.71924e-3, -0.4900e-4 }
         };
 
-        const Evaluation& theta = temperature - 273.15;
+        const Evaluation theta = temperature - 273.15;
 
         Evaluation S = salinity;
-        const Evaluation& S_lSAT =
+        const Evaluation S_lSAT =
             f[0]
             + f[1]*theta
             + f[2]*pow(theta, 2)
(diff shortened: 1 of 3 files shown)


akva2:
jenkins build this please

opm/material/components/Brine.hpp
```diff
@@ -149,22 +149,22 @@ class Brine : public Component<Scalar, Brine<Scalar, H2O> >
...
-        static const Scalar a[4][3] = {
+        static constexpr Scalar a[4][3] = {
             { -9633.6, -4080.0, +286.49 },
             { +166.58, +68.577, -4.6856 },
             { -0.90963, -0.36524, +0.249667e-1 },
             { +0.17965e-2, +0.71924e-3, -0.4900e-4 }
         };
 
-        const Evaluation& theta = temperature - 273.15;
+        const Evaluation theta = temperature - 273.15;
```
blattms, line 164: Now I'm curious, why  is this a dangling reference. Shouldn't the lifetime be extended until theta goes out of scope?
That being  said, I don't oppose the change as the performance impacted is expected to be minimal.

akva2: uhm, you are correct. those commits are just noise, not sure where my head was at. i can remove them.

akva2: i am however quite certain they do no harm, as surely these will be moved.

PR #523, opened 2022-08-02
Cleanups and modernization in pvt classes

akva2:


akva2:
jenkins build this please

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
```diff
@@ -706,7 +704,7 @@ class WetGasPvt
 private:
     void updateSaturationPressure_(unsigned regionIdx)
     {
-        typedef std::pair<Scalar, Scalar> Pair;
+        using Pair = std::pair<Scalar, Scalar>;
```
blattms, line 707: Can't help it as the regular person for this is not here :wink:.
Do we even need this? What about just using this below
```
-            Pair val(Rv, pSat);
-            pSatSamplePoints.push_back(val);
+            pSatSamplePoints.emplace_back(Rv, pSat);
```

opm/material/fluidsystems/blackoilpvt/WetHumidGasPvt.hpp
```diff
@@ -828,7 +826,7 @@ class WetHumidGasPvt
 private:
     void updateSaturationPressure_(unsigned regionIdx)
     {
-        typedef std::pair<Scalar, Scalar> Pair;
+        using Pair = std::pair<Scalar, Scalar>;
```
blattms, line 829: skip and use emplace_back, below?

opm/material/fluidsystems/blackoilpvt/WetGasPvt.hpp
akva2: okay. Pair is also used in lambda parameters below, so i changed to a variadic lambda there.

opm/material/fluidsystems/blackoilpvt/WetHumidGasPvt.hpp
akva2: ditto.

akva2:
jenkins build this please

PR #525, opened 2022-08-03
Allow private pvt templates

akva2:
See https://github.com/OPM/opm-simulators/pull/3987 for rationale


akva2:
jenkins build this please

PR #528, opened 2022-08-15
changed: consistently use stack/C arrays

akva2:
This is something I found while doing my private templates experiments.

Stack arrays are used in most of the code. Not using them here means we are instancing potentially hundreds of extra templates for very little reason.


akva2:
jenkins build this opm-simulators=4001 please

blattms:
Now I am completely confused. Seems to be orthogonal to #529

akva2:
exactly. we were testing both to see if it has runtime implications.

blattms:
WIP?

PR #531, opened 2022-09-01
Add enthalpy and internalEnergy to co2brinepvt

totto82:


totto82:
jenkins build this please

bska (approved):
This is okay but if we're going to continue to develop this utility then we should probably refactor the nested control structures a bit to simplify future maintenance.  For now I'll merge into master.

PR #533, opened 2022-09-20
changing assertions to throw with information output for function findSegmentIndex_ of class Tabulated1DFunction

GitPaean:


GitPaean:
jenkins build this please

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -35,6 +35,7 @@
 #include <iostream>
 #include <tuple>
 #include <vector>
+#include "fmt/format.h"
```
akva2, line 38: Is this used in tests in this module? Internal libfmt is not available in this repo.

GitPaean: sorry, did not know about that. Should we include Internal libfmt here or I can change the code without using `fmt::format`?

Either way will work for me.

akva2: I think for now it is better to use an sstream than go through the hassle of adding internal libfmt. Hopefully we will merge this repo with opm-common soon, atleast that has is the plan, and then the problem goes away..

GitPaean: Sure. I will do that.

GitPaean:
jenkins build this please

akva2 (approved):

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -475,8 +482,27 @@ class Tabulated1DFunction
                     lowerIdx = pivotIdx;
             }
 
-            assert(xValues_[lowerIdx] <= x);
-            assert(x <= xValues_[lowerIdx + 1]);
+            if (xValues_[lowerIdx] > x || x > xValues_[lowerIdx + 1]) {
+                std::ostringstream sstream;
+                sstream << "Problematic interpolation/extrapolation segment is found for the input value " << Opm::getValue(x)
+                        << "\nthe lowe index of the segment is " << lowerIdx << ", the size of the table is " << numSamples()
```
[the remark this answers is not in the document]
GitPaean: Thanks. Will provide a PR for that, maybe other small things all together.

opm/material/common/Tabulated1DFunction.hpp
```diff
@@ -475,8 +482,27 @@ class Tabulated1DFunction
                     lowerIdx = pivotIdx;
             }
 
-            assert(xValues_[lowerIdx] <= x);
-            assert(x <= xValues_[lowerIdx + 1]);
+            if (xValues_[lowerIdx] > x || x > xValues_[lowerIdx + 1]) {
```
GitPaean, line 485: For the scenarios that we are trying to handling, x is a `NaN`.  Then the behavior here is a little different from the original assertions trying to do. 

When `NaN` is involved, all (almost) the logic comparisons are `false`. As a result, in the original code, assertion failures will be triggered but in the new code in this PR, it will not throw. 

Luckily, the new PR https://github.com/OPM/opm-material/pull/534 complements the check for `NaN` values.

GitPaean: or 
```c++
if ( !(xValues_[lowerIdx] <= x && x <= xValues_[lowerIdx + 1]) ) {

```
will also do the work.

PR #535, opened 2022-10-07
Killough's hysteresis model for relperm and cap-pressure.

osae:
Functionality related to kw `EHYSTR`, item#2 (flags 2 and 3) and item#5 (flags PC and BOTH). 
Related pull requests: https://github.com/OPM/opm-common/pull/3170 and https://github.com/OPM/opm-simulators/pull/4148.

### DIFF
--- a/opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
+++ b/opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
@@ -116,6 +116,9 @@ public:
 
         oilWaterSystem_ = (twoPhaseSystem == EclOilWaterSystem);
 
+        if (!config().enableHysteresis())
+            return;
+
         if (config().krHysteresisModel() == 2 || config().krHysteresisModel() == 3 || config().pcHysteresisModel() == 0) {
             if (twoPhaseSystem == EclGasOilSystem) {
                 Sncrd_ = info.Sgcr+info.Swl;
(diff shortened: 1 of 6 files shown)


opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
```diff
@@ -241,11 +339,20 @@ class EclHysteresisTwoPhaseLawParams : public EnsureFinalized
     void update(Scalar pcSw, Scalar /* krwSw */, Scalar krnSw)
     {
         bool updateParams = false;
-        if (pcSw < pcSwMdc_) {
+
+        if (config().pcHysteresisModel() == 0 && pcSw < pcSwMdc_) {
+            if (pcSwMdc_ == 2.0 && pcSw+1.0e-6 < Swcrd_ && oilWaterSystem_) {
+               initialImb_ = true;
```
totto82, line 345: Are the results sensitiv to the 1.0e-6 here. I notice several hard-coded values like 1.0e-12 and 1.0e-8. It is a bit hard to evaluate the significance of these numbers.

totto82 (commented):
Thanks for work. I think this can be merged. But I would like your comments on the sensitivity of the hard-coded small value.

opm/material/fluidmatrixinteractions/EclHysteresisTwoPhaseLawParams.hpp
osae: Sensitive: Yes, behavior will in general be different if we consider the process to be initial imbibition (following pure imb curve) compared to assuming reversal from drainage (following scanning curve). Whether the 1e-6 term is reasonable, or more general how to make the distinction between the two cases, can of course be debated.

Significance of 1e-N: Except from the case discussed above, the 1e-6 and 1e-8 are used to guard against division by zero, suggest to use 1e-8 everywhere. The 1e-12 is used once, and (as far as I remember) 1e-6**2 was chosen after some indications that 1e-6 was too large here.

totto82: Thanks for the comment. I am ok with the current values.

PR #547, opened 2022-12-05
hmm. this should be set

akva2:
debugging.. testing effect of this obviously missing define. should only affect tests but.


akva2:
jenkins build this please

